PR #12, opened 2022-05-24
feat: port over err compare test util

yolocs:
Copied from: https://github.com/abcxyz/lumberjack/blob/main/clients/go/pkg/errutil/errutil.go


2142838352217396206439175:
cioei thanks for the invitation U2.0wik

2142838352217396206439175:
Eddwsworld ahiorv ir no 20?!

PR #19, opened 2022-06-17
feat: add fake grpc server testutil

yolocs:
Copied from: https://github.com/abcxyz/lumberjack/blob/main/clients/go/pkg/testutil/grpc.go

### DIFF
--- a/testutil/grpc.go
+++ b/testutil/grpc.go
@@ -1,0 +1,54 @@
+// Copyright 2022 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


testutil/grpc.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+// FakeGRPCServer creates and registers a fake grpc server for testing. It
+// returns the bound address and a connected client. It ensures the server is
+// stopped when tests finish. It returns an error if the connection does not
+// establish.
+func FakeGRPCServer(tb testing.TB, registerFunc func(*grpc.Server)) (string, *grpc.ClientConn) {
```
sethvargo, line 29: Since we're pulling this into its own package with a public API, we should extract `register` func as a publicly-registered type.

sethvargo (approved):

testutil/grpc.go
yolocs: Why?

sethvargo: It's better for documentation (godoc), so we can cleanly call out the invariants of the input function independent of the function definition itself (example http.HandlerFunc).

sethvargo: It's not blocking for this (I approved the PR), but I think it would be nice for the godoc.

yolocs: Done

Issue #20, opened 2022-06-21
bug: validate JWTs


raserva:
Right now, the code for getting principals from JWTs does no validation on them. We should be pulling the public keys in order to validate that they have been signed correctly and are valid.

sethvargo:
This seems bad

raserva:
I agree. After Chen's comments, I may not be moving this code over, but in lumberjack this absolutely should be updated

PR #21, opened 2022-06-21
feat: relocate principal extraction code to pkg

raserva:
We will need principal extraction for Lumberjack, JVS, and probably future services as well. I've copied over the existing code from Lumberjack ( https://github.com/abcxyz/lumberjack/blob/main/clients/go/pkg/security/grpc.go#L42 , https://github.com/abcxyz/lumberjack/blob/main/clients/go/apis/v1alpha1/config.go#L267 ) with minor modifications 

- renamed struct to be more obvious, i found the name FromRawJWT to be a bit confusing, especially as it was used to name both an individual JWT rule and a group of rules
- Added TODO for JWT validation. Will follow up in another PR, but the existing code does no validation of the JWTs. https://github.com/abcxyz/pkg/issues/20

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,9 +4,10 @@ go 1.18
 
 require (
 	cloud.google.com/go/compute v1.6.1
+	github.com/golang-jwt/jwt v3.2.2+incompatible
 	github.com/google/go-cmp v0.5.8
 	go.uber.org/zap v1.21.0
-	google.golang.org/grpc v1.46.2
+	google.golang.org/grpc v1.47.0
 )
 
 require (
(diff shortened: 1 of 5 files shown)


raserva:
I left the config-related stuff in (ex: `yaml:"key,omitempty"`) so that these could be used as drop-in replacements (other than one yaml name). If we prefer to keep them separate, i can remove the config stuff, and have lumberjack/jvs/etc specify their own config that lines up with this.

yolocs:
I have some different thoughts. We added JWT rules in lumberjack to enable clients configurability but it makes less sense elsewhere. A more common util is 

```go
// Assume the package is named as jwtutil.
// Retrieve the principal from the "email" claim. 
jwtutil.EmailPrincipal(token string) (string, error) { ... }
```

And this function can be reused in lumberjack. It's pretty trivial to get value from grpc metadata. For the services we implement, we can hardcode where the token is from (e.g. Authorization header as a bearer token).

Later on, we can add jwt validation util to the `jwtutil` package.

Thoughts?

raserva:
> If you really want to make it work for gRPC. I'd suggest an interceptor. Overtime we can build up common middlewares for grpc.

The JVS is a gRPC service, so we will need gRPC support. However, i agree with your above comment - supporting this level of configurability is probably unnecessary - we can just assume callers are using iam and have a method to retrieve the principal from iam token.

bradegler (approved):

PR #25, opened 2022-06-27
feat: switched to option pattern

raserva:
follow up from https://github.com/abcxyz/pkg/pull/24

### DIFF
--- a/grpcutil/jwt_auth.go
+++ b/grpcutil/jwt_auth.go
@@ -39,22 +39,36 @@ type JWTAuthenticationHandler struct {
 	// the jwt ourselves. This flag skips the validation, and instead trusts the JWT has been validated
 	// upstream.
 	ValidationDisabled bool
+	Endpoint           string
 }
 
 // NewJWTAuthenticationHandler returns a JWTAuthenticationHandler with a verifier initialized. Uses defaults
 // for JWT related fields that will retreive a user email when using IAM on GCP.
-func NewJWTAuthenticationHandler(ctx context.Context, endpoint string, disableValidation bool) (*JWTAuthenticationHandler, error) {
-	verifier, err := jwtutil.NewVerifier(ctx, endpoint)
-	if err != nil {
-		return nil, err
-	}
-	return &JWTAuthenticationHandler{
-		Verifier:           verifier,
+func NewJWTAuthenticationHandler(ctx context.Context, opts ...HandlerOption) (*JWTAuthenticationHandler, error) {
+	j := &JWTAuthenticationHandler{
 		JWTPrefix:          "bearer ",
 		JWTKey:             "authorization",
 		PrincipalClaimKey:  "email",
-		ValidationDisabled: disableValidation,
-	}, nil
+		ValidationDisabled: false,
+	}
+	for _, opt := range opts {
+		opt(j)
+	}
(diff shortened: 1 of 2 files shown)


grpcutil/jwt_auth.go
```diff
@@ -39,22 +39,36 @@ type JWTAuthenticationHandler struct {
...
+	j := &JWTAuthenticationHandler{
 		JWTPrefix:          "bearer ",
 		JWTKey:             "authorization",
 		PrincipalClaimKey:  "email",
-		ValidationDisabled: disableValidation,
-	}, nil
+		ValidationDisabled: false,
+	}
+	for _, opt := range opts {
+		opt(j)
```
sethvargo, line 55: This will panic if opt is `nil` (which is possible since nil will satisfy that interface)

grpcutil/jwt_auth.go
```diff
@@ -108,3 +122,40 @@ func (g *JWTAuthenticationHandler) RequestPrincipal(ctx context.Context) (string
 
 	return principal, nil
 }
+
+type HandlerOption func(handler *JWTAuthenticationHandler)
```
sethvargo, line 126: The option pattern usually _returns_ a pointer too. That gives opts the flexibility to possibly return an entirely new object.

sethvargo (approved):

grpcutil/jwt_auth.go
```diff
@@ -39,22 +39,39 @@ type JWTAuthenticationHandler struct {
...
 // NewJWTAuthenticationHandler returns a JWTAuthenticationHandler with a verifier initialized. Uses defaults
 // for JWT related fields that will retreive a user email when using IAM on GCP.
-func NewJWTAuthenticationHandler(ctx context.Context, endpoint string, disableValidation bool) (*JWTAuthenticationHandler, error) {
-	verifier, err := jwtutil.NewVerifier(ctx, endpoint)
-	if err != nil {
-		return nil, err
-	}
-	return &JWTAuthenticationHandler{
-		Verifier:           verifier,
+func NewJWTAuthenticationHandler(ctx context.Context, opts ...HandlerOption) (*JWTAuthenticationHandler, error) {
```
yolocs, line 48:
```suggestion
func NewJWTAuthenticationHandler(ctx context.Context, opts ...JWTAuthOption) (*JWTAuthenticationHandler, error) {
```

I recommend rename the option type to be more specific to this handler. This would help avoid future naming conflicts with other utils in the package.

grpcutil/jwt_auth.go
```diff
@@ -108,3 +125,45 @@ func (g *JWTAuthenticationHandler) RequestPrincipal(ctx context.Context) (string
 
 	return principal, nil
 }
+
+type HandlerOption func(handler *JWTAuthenticationHandler) *JWTAuthenticationHandler
+
+// NoValidation disables certificate validation for JWT.
+func NoValidation() HandlerOption {
```
yolocs, line 132: Similarly, I recommend add `JWTAuth` as a substring for these functions. E.g. `NoJWTAuthValidation`, `WithJWTAuthEndpoint`. This would help avoid future naming conflicts.

yolocs (changes requested):
Naming

PR #28, opened 2022-07-18
Add auto-expiration to the cache implementation

sethvargo:
The original cache implementation was inspired by the Exposure Notifications cache implementation, which specifically didn't have a requirement for expiring cache entires because public keys were rotated infrequently and the services restart frequently. Since this is a general-purpose TTL cache now, we need expiration.

Introducing expiration means we need a background goroutine to manage expiration. This necessitates the introduction of a `Stop()` function to terminate the cacher when the system is shutting down. It also requires some more careful locking around when sweeping occurs.

To minimize lock contention, this implementation uses a linked list to keep track of cache items as they are inserted into the cache. New items are inserted at the end of the list, and the sweep operation reads from the beginning until it reaches a non-expired item. This avoids needing to read every item in the cache map (which is unordered) at the cost of additional memory overhead to maintain the order. Since we already have to allocate onto the heap for the cache, the linked list is primarily pointers instead of values.

Other high-level changes:

-   `Set()` no longer returns an error. It never should have.
-   Internal storage uses `time.Time` instead of an int to account for clock skew.


cache/cache.go
```diff
@@ -61,76 +63,97 @@ func New[T any](expireAfter time.Duration) *Cache[T] {
...
-// If there is a race, and the item has been refreshed, it will not be purged.
-func (c *Cache[T]) purgeExpired(name string, expectedExpiryTime int64) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
 
-	if item, ok := c.data[name]; ok && item.expiresAt == expectedExpiryTime {
-		// found, and the expiry time is still the same as when the purge was requested.
-		delete(c.data, name)
+	// Start the sweep, with a minimum sweep of 50ms.
+	sweep := expireAfter / 4.0
```
raserva, line 73: Why a quarter of expireAfter? I do understand why we would want to do sweeps more quickly than the expiry time, although i'm curious why specifically its a quarter. Is this just a number picked out of the air?

cache/cache.go
```diff
@@ -61,76 +63,97 @@ func New[T any](expireAfter time.Duration) *Cache[T] {
...
+	if min := 50 * time.Millisecond; sweep < min {
```
raserva, line 74: Same question about why 50 ms. I would expect, for example, a cache of public keys could get away with a longer time without issue.

cache/cache.go
```diff
@@ -61,76 +63,97 @@ func New[T any](expireAfter time.Duration) *Cache[T] {
...
-// Clear removes all items from the cache, regardless of their expiration.
+// Clear removes all items from the cache, regardless of their expiration. Note
+// this is different from Stop() which deletes all cached items and prevents new
+// items from being added.
 func (c *Cache[T]) Clear() {
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	c.data = make(map[string]item[T])
+
+	if c.isStopped() {
```
raserva, line 101: We have a lot of handling for stopping the cache. I'm a bit curious the use case we expect - i can't think of cases where i'd want to stop the cache without shutting down my application.

cache/cache.go
```diff
@@ -38,21 +39,22 @@ type Func[T any] func() (T, error)
 // For performance, it's strongly recommended that you store pointers to objects
 // instead of actual objects.
 type Cache[T any] struct {
-	data        map[string]item[T]
+	// data is the actual internal cache storage.
+	data map[string]*cacheListItem[T]
+
+	// head points to the head of the linked list, tail points to the tail.
+	head, tail *cacheListItem[T]
```
raserva, line 46: So, if we're trying to make a generalizable cache that is time-based, i question using a linked list. That is the typical implementation for LRUs, so i understand why it was added. But i would expect that in a truly generalizable time-based cache implementation, i would be able to specify different expiration times for different pieces of data. We don't necessarily have to implement that now, but if we wanted to support that, it wouldn't fit will with the linked-list based implementation, which would assume that items are in order of expiration.

raserva (commented):
I mentioned this over chat - but i think its worth documenting why we're creating our own cache, rather than using an existing one.

cache/cache.go
```diff
@@ -139,40 +162,163 @@ func (c *Cache[T]) WriteThruLookup(name string, primaryLookup Func[T]) (T, error
...
+
+	c.set(name, object, now)
+}
+
+func (c *Cache[T]) set(name string, object T, now time.Time) {
+	// Calculate expiration after acquiring a lock. The item is valid from when
+	// insertion started, not when insertion finishes.
+	exp := now.Add(c.expireAfter)
+
+	node, ok := c.data[name]
```
mikehelmick, line 208: There is a bug here - if the value is found, the node that is being pointed to is reused but not _moved_ in the list. 

This actually breaks the linked list chain an keeps it at it's current spot, blocking further cleanup.

mikehelmick (changes requested):

cache/cache.go
sethvargo: It gives us a margin of error above 1/2, but doesn't put crazy pressure on the CPU for small expiration times.

cache/cache.go
sethvargo: I expect most caches to be O(minutes) or O(hours). This is guarding against a case where someone makes a cache of 1ns and then we expect the sweep to happen more often than is physically possible. Even for small values (e.g. 10ms), we want to avoid pegging the CPU. The cache still checks if items are expired before returning them, so there's a guard against returning an item that should have been expired.

cache/cache.go
sethvargo: Yea, you should really only terminate the cache if you're:

1. shutting down
2. creating an entirely new cache (e.g. to change the TTL)

Without this logic, you'd get really difficult to diagnose NPEs since we want to zero out as much of the old cache as possible to enable efficient GC. While still a panic, this is a consistent and debuggable panic.

cache/cache.go
sethvargo: Yes, if we went with a generic TTL-based cache where individual items could have their own TTLs, we'd probably switch to a heap to make walking more efficient. I don't think we need that optimization here yet.

cache/cache.go
```diff
@@ -61,76 +63,97 @@ func New[T any](expireAfter time.Duration) *Cache[T] {
...
+	return c
 }
 
-// Size returns the number of items in the cache.
+// Size returns the current number of items in the cache.
 func (c *Cache[T]) Size() int {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
+
+	if c.isStopped() {
```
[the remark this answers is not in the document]
sethvargo: All these checks are pre-subposed on the mutex being held, so I don't think that race can actually happen.

cache/cache.go
```diff
@@ -139,40 +162,163 @@ func (c *Cache[T]) WriteThruLookup(name string, primaryLookup Func[T]) (T, error
...
 
-	c.data[name] = item[T]{
-		object:    object,
-		expiresAt: time.Now().Add(c.expireAfter).UnixNano(),
+	if !atomic.CompareAndSwapUint32(&c.stopped, 0, 1) {
+		return
+	}
+	close(c.stopCh)
+
+	for k, v := range c.data {
```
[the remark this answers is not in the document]
sethvargo: Yup lol

cache/cache.go
sethvargo: Addressed, with many more tests.

cache/cache.go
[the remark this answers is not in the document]
sethvargo: De-duped.

cache/cache.go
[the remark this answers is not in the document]
sethvargo: I like comments.

cache/cache.go
NishThakkar: I recommend adding the "why" as a code comment, so someone years later knows the context and can make a change if needed.

cache/cache.go
NishThakkar: I recommend adding the "why" as a code comment, so someone years later knows the context and can make a change if needed.

raserva (approved):

cache/cache.go
sethvargo: Commented inline

cache/cache.go
sethvargo: Commented inline

PR #32, opened 2022-09-13
Fix missing docstrings, improve test coverage, and improve error mess…

drevell:
…age for nonexistent mysql version

### DIFF
--- a/mysqltest/config.go
+++ b/mysqltest/config.go
@@ -24,6 +24,8 @@ type config struct {
 	progressLogger Logger
 }
 
+// Logger allows the caller to optionally provide a custom logger for printing status updates about
+// MySQL startup progress. The default is to use the go "log" package.
 type Logger interface {
 	Printf(fmtStr string, args ...any)
 }
@@ -44,6 +46,8 @@ func buildConfig(opts ...Option) *config {
 	return config
 }
 
+// Option sets a configuration option for this package. Users should not implement these functions,
+// they should use one of the With* functions.
 type Option func(*config) *config
 
 // WithKillAfterSeconds is an option that overrides the default time period after which the mysql docker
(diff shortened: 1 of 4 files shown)


mysqltest/config.go
```diff
@@ -68,6 +72,15 @@ func WithVersion(v string) Option {
 	}
 }
 
+// WithLogger overrides the default logger. This logger will receive messages about MySQL startup
+// progress. The default is to use the go "log" package.
+func WithLogger(l Logger) Option {
```
sethvargo, line 77: No action, but you should check out jba's slog proposal on the Go repo if you're interested for structured logging in the stdlib.

mysqltest/mysqltest_test.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	if conf.mySQLVersion != "2" {
+		t.Errorf(`got mySQLVersion=%v", want "2"`, conf.mySQLVersion)
+	}
+	if _, ok := conf.progressLogger.(*fakeLogger); !ok {
+		t.Errorf("got progressLogger type %T, want %T", conf.progressLogger, logger)
+	}
+}
+
+type fakeLogger struct{}
```
sethvargo, line 64: For testing, we might want to create a `testLogger` that calls `t.Log`. That way we see the log output on test failures:

```go
type testLogger struct {
  tb testing.TB
}

func (l *testLogger) Printf(msg string, args ...any) {
  l.tb.Logf(msg, args...)
}
```

Then &testLogger{t} should just work.

sethvargo (approved):

yolocs (approved):
Side feature request: I wonder if it's reasonable to allow images not on docker registry (e.g. GCR mirror). E.g. GCR/AR probably have better reliability.

mysqltest/config.go
drevell: Thanks, I hadn't seen that but I welcome it. https://github.com/golang/go/discussions/54763 for posterity.

mysqltest/mysqltest_test.go
drevell: Done, thanks for the idea.

drevell:
> Side feature request: I wonder if it's reasonable to allow images not on docker registry (e.g. GCR mirror). E.g. GCR/AR probably have better reliability.

Hmm good point, I hadn't given much thought to reliability of docker image downloads.

PR #33, opened 2022-09-15
feat: add config loader util

yolocs:


cfgloader/cfgloader.go
```diff
@@ -0,0 +1,114 @@
...
+// The given config type must have the yaml tags to load from yaml bytes.
+// It must have the [env tag] to load from env vars. E.g.
+//
+//	type Cfg struct {
+//		StrVal string `yaml:"str_val,omitempty" env:"STR_VAL,overwrite,default=foo"`
+//		NumVal int    `yaml:"num_val,omitempty" env:"NUM_VAL,overwrite,default=1"`
+//	}
+//
+// [env tag]: https://github.com/sethvargo/go-envconfig
+func Load(ctx context.Context, cfg any, opt ...Option) error {
```
capri-xiyue, line 81: Do we need to enforce `cfg` to implement the Validatable interface and use Validatable as the type of cfg instead of any?

yolocs: No, we don't have to ask every config to implement `Validate` func. At the bottom of the func, I assert the interface and only call `Validate` if implemented.

capri-xiyue: ack

sethvargo:
So, I think this is fine. But in reality, each project should probably define it's own config via `NewConfig` that has defaults. It's fine if those functions call envconfig to produce the defaults, but it seems weird to centralize this in pkg.

yolocs:
> So, I think this is fine. But in reality, each project should probably define it's own config via `NewConfig` that has defaults. It's fine if those functions call envconfig to produce the defaults, but it seems weird to centralize this in pkg.

As long as we have a consistent way to set defaults, I'm good. At the moment, pretty much every config we have uses `envconfig` - for one we need it to load overwrites from env vars, for two we use it to set defaults. This makes me think it might be a good idea to have the loading logic as a common functionality.

sethvargo:
The thing I'm concerned with is that we might be conflating two things.

1. **Default config** - this should be "constant" within a versioned release regardless of the environment. In other words, we should _only_ care about the `default=` part of envconfig. We could accomplish this by processing the struct with a `nil` `MapLookuper`.
2. **Compiled config** - this should be the callers' specific config, based on the environment + falling back to the defaults if unspecified.

sqin2019 (approved):

yolocs:
Let me discuss this with you offline. I don't immediately see why they're conflating.

sethvargo (approved):

yolocs:
@sethvargo Clarified usage as we discussed.

PR #34, opened 2022-09-22
Update all dependencies to their latest versions

sethvargo:


.github/workflows/go-lint.yml
```diff
@@ -70,7 +70,6 @@ jobs:
               "bidichk",
               "bodyclose",
               "containedctx",
-              "deadcode",
```
sethvargo, line 73: These are deprecated and replaced by "unused" (which we already have).

drevell (approved):

PR #35, opened 2022-09-23
Add helper for making a logger tied to a test context

sethvargo:
This adds a new function to the logging package that enables us to couple the logger to the testing context. This will clean up test output and ensure we only see log messages relevant to the failed test(s).

### DIFF
--- a/logging/logger.go
+++ b/logging/logger.go
@@ -26,6 +26,7 @@ import (
 	"cloud.google.com/go/compute/metadata"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest"
 	"google.golang.org/grpc"
 	grpcmetadata "google.golang.org/grpc/metadata"
 )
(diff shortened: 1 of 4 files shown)


logging/logger.go
```diff
@@ -106,6 +107,20 @@ func FromContext(ctx context.Context) *zap.SugaredLogger {
...
+// information if specific test fails or is run in verbose mode. See [zaptest]
+// for more information.
+//
+//	func TestMyThing(t *testing.T) {
+//		logger := logging.TestLogger(t)
+//		thing := &MyThing{logger: logger}
+//	}
+//
+// [zaptest]: https://pkg.go.dev/go.uber.org/zap/zaptest
+func TestLogger(tb zaptest.TestingT) *zap.SugaredLogger {
```
drevell, line 120: Optional: consider accepting a parameter "opts ...zaptest.LoggerOption" to this function, and passing them through to zaptest.NewLogger. That would let the caller override the level. Dunno if we actually care about that though.

PR #38, opened 2022-09-26
feat: allow cfgloader to set defaults with code

yolocs:


sethvargo (approved):

cfgloader/cfgloader.go
```diff
@@ -105,6 +112,10 @@ func Load(ctx context.Context, cfg any, opt ...Option) error {
 		return fmt.Errorf("failed to load config: %w", err)
 	}
 
+	if d, ok := cfg.(Defaultable); ok {
+		d.SetDefault()
```
drevell, line 116: I'm probably missing the point, but I would have thought that the default would be set first, then overwritten. As I see it here, the defaults come last. This works for things can can be nil-checked like maps and pointers, but for an integer variable, the SetDefault function can't tell whether it's 0 because it was unset or because the environment specifically requested a value of 0.

cfgloader/cfgloader.go
```diff
@@ -105,6 +112,10 @@ func Load(ctx context.Context, cfg any, opt ...Option) error {
 		return fmt.Errorf("failed to load config: %w", err)
 	}
 
+	if d, ok := cfg.(Defaultable); ok {
```
pjh, line 115: I think all of your new test cases test the positive path here (cfg implements Defaultable), right? If you want to be thorough you could test the negative path also.

pjh (approved):

cfgloader/cfgloader.go
```diff
@@ -71,12 +78,12 @@ func WithLookuper(lookuper envconfig.Lookuper) Option {
 //
 // The values loaded later will overwrite previously loaded values.
 //
-// The given config type must have the yaml tags to load from yaml bytes.
-// It must have the [env tag] to load from env vars. E.g.
+// The given config type must have the yaml tags to load from yaml bytes. It
```
drevell, line 81: Do the docs at line 75 need to be updated to explain the new default behavior?

drevell (changes requested):

cfgloader/cfgloader.go
yolocs: Good point. Let me add.

cfgloader/cfgloader.go
yolocs: You're totally right.

cfgloader/cfgloader.go
yolocs: Separated validatable, defaultable tests

cfgloader/cfgloader.go
```diff
@@ -96,6 +105,10 @@ func Load(ctx context.Context, cfg any, opt ...Option) error {
 		}
 	}
 
+	if d, ok := cfg.(Defaultable); ok {
+		d.SetDefault()
```
drevell, line 109: It's a little surprising to me that the defaults override the YAML. Have I understood correctly?

drevell (changes requested):

cfgloader/cfgloader.go
yolocs: That really depends on the `SetDefault()` logic. One should check zero values (ack the limitation on primitive types but it's not a major concern) before setting a default.

drevell: SGTM, I'll go ahead and approve.

I'll continue making my feedback, which you can decide whether it makes sense to incorporate. If the SetDefault() call was first (before yaml), then the implementation of SetDefault would be impossible to do wrong. SetDefault could just blindly set the default values in each field, without any risk of the user forgetting or misunderstanding the need for nil checks.

drevell (approved):

cfgloader/cfgloader.go
yolocs: Appreciate the feedback!

I think your suggestion would work well except when the `SetDefault` is not self-contained, e.g. it needs to look at other config attributes and decide a default. E.g. In lumberjack we have a config to indicate where to fetch the ID token from the grpc metadata. We only want to set the default if the parent block is not nil.

yolocs: I ack that the current approach is not perfect. When we get the value from env var but we have `SetDefault` before loading the overwrites from env var. Maybe that's why I added `SetDefault` at last...

drevell: Ah, I see now, thanks for explaining.

yolocs:
Had an offline discussion with @drevell. With `SetDefault`, there is a dilemma - We want it to set defaults only after the values are loaded from various places (original struct, yaml, env vars). It's more like pre-validation processing but in essence it's still about setting defaults. To avoid confusion, if one wants to have such pre-validation processing, they can do it in the `Validation()` func instead. The additional interface here was not thought through. Thus, dropping this change.

sethvargo:
> To avoid confusion, if one wants to have such pre-validation processing, they can do it in the Validation() func instead.

That's fine and makes sense, but we should update the godoc and interface description to call this out explicitly. Something like "In addition to validating inputs, [Validate] may optionally set defaults.". @yolocs @drevell

PR #40, opened 2022-10-07
Add generic worker for parallelizing tasks

sethvargo:


worker/worker.go
```diff
@@ -0,0 +1,164 @@
...
+// it was enqueued, but is not guaranteed to complete in the order in which it
+// was enqueued (i.e. this is not a pipeline).
+//
+// If the provided concurrency is less than 1, it defaults to the number of CPU
+// cores.
+func New[T any](concurrency int64) *Worker[T] {
+	if concurrency < 1 {
+		concurrency = int64(runtime.NumCPU())
+	}
+	if concurrency < 1 {
```
drevell, line 74: I think there might be a copy-paste mistake here, there are two checks for "concurrency < 1"

worker/worker.go
```diff
@@ -0,0 +1,164 @@
...
+		concurrency = 1
+	}
+
+	return &Worker[T]{
+		size:    concurrency,
+		i:       -1,
+		sem:     semaphore.NewWeighted(concurrency),
+		results: make([]*result[T], 0, concurrency*4),
```
drevell, line 82: The meaning of "4" isn't immediately obvious, could you add a named const or a comment?

worker/worker_doc_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	results, err := w.Done(ctx)
+	if err != nil {
+		// TODO: check err
+	}
+	_ = results
+}
+
+func Example_hTTP() {
```
drevell, line 48: hTTP -> HTTP

worker/worker_test.go
```diff
@@ -0,0 +1,152 @@
...
+		want := []int{0, 1, 2, 3, 4}
+		var got []int
+		for _, result := range results {
+			got = append(got, result.Value)
+		}
+		if diff := cmp.Diff(want, got); diff != "" {
+			t.Errorf("justs: diff (-want, +got):\n%s", diff)
+		}
+	})
+}
```
drevell, line 152: Suggested tests:
 
 - Canceled context with active workers and a blocked enqueuer thread

Extra credit:

 - Multiple goroutines enqueueing work, and run the test with the race detector. Maybe this is excessive because this won't be used in this way.

drevell (approved):

worker/worker.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// Do adds new work into the queue. If there are no available workers, it blocks
+// until a worker becomes available or until the provided context is cancelled.
+// The function returns when the work has been successfully scheduled.
+//
+// To wait for all work to be completed and read the results, call [Done]. After
+// the worker has been stopped via [Done], this will return [ErrStopped].
+//
+// Never call [Do] from within a [Do] function because it will deadlock.
```
yolocs, line 93: Can we add some doc about how to handle error returned from `Do`? Should I retry? Can I continue to call `Done` to wait? Or cancel everything (via context)? Or else?

worker/worker_test.go
yolocs: +1 testing context cancel

worker/worker.go
sethvargo: This is intended. The first is checking if the incoming value is < 1, in which case it defaults to the number of CPUs. The second check is ensuring that we always have at least one concurrency (if for some reason NumCPU returned 0).

worker/worker_doc_test.go
sethvargo: I agree this is super annoying, but it's actually intentional. It can't be `HTTP` because it has to start with a lowercase letter. If I make it `_http`, then the example name looks silly:

screenshot-20221007-rgQkooBa@2x

Here's how it looks as-written:

screenshot-20221007-D3nMJ2C7@2x

worker/worker.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	if err := w.sem.Acquire(ctx, 1); err != nil {
+		return err
+	}
+
+	// It's possible the worker was stopped while we were waiting for the
+	// semaphore to acquire, but the worker is actually stopped.
+	if w.isStopped() {
+		defer w.sem.Release(1)
```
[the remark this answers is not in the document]
sethvargo: This is necessary. We acquire the semaphore above, but `Done` might be waiting to acquire all semaphores. If we don't release, it's a deadlock.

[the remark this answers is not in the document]
yolocs: I don't mean the `Release`. Just the `defer`

worker/worker.go
drevell: Ah, now I see.

worker/worker.go
sethvargo: Done.

sethvargo:
@drevell @yolocs added more tests and docs - PTAL

worker/worker_test.go
```diff
@@ -0,0 +1,228 @@
...
+		var got []int
+		for _, result := range results {
+			got = append(got, result.Value)
+		}
+		if diff := cmp.Diff(want, got); diff != "" {
+			t.Errorf("justs: diff (-want, +got):\n%s", diff)
+		}
+	})
+
+	t.Run("concurreny", func(t *testing.T) {
```
drevell, line 199: "concurreny"

drevell (approved):

yolocs (approved):

PR #47, opened 2022-11-09
Set up a reusable action to setup necessary tools. For ex: skaffold

agarwalnit:
As we are setting up e2e test environment for Verily using minikube and skaffold. Currently, a test environment setup requires setting up multiple tools, for ex: minikube, skaffold. 

Creating a reusable action which can be used by other Verily teams to directly use and simplify their github workflows. 

TESTED=Tested by pushing to a branch and referencing in an existing github workflow for verily : https://github.com/verily-src/ss-api-e2e-tests/actions/runs/3425619554/jobs/5706583793

### DIFF
--- a/.github/actions/setup-e2e-tools/action.yml
+++ b/.github/actions/setup-e2e-tools/action.yml
@@ -1,0 +1,13 @@
+name: 'setup-e2e-tools'
+description: 'Set up required tooling for creating an e2e test environment'
+
+runs:
+  using: "composite"
+  steps:
+    - name: 'Install Skaffold'
+      shell: bash
+      run: |
+        curl -Lo skaffold https://storage.googleapis.com/skaffold/releases/latest/skaffold-linux-amd64 && sudo install skaffold /usr/local/bin/ && skaffold version
+    - name: 'Install and start minikube'
+      id: minikube
+      uses: medyagh/setup-minikube@b127ace2ffdece4eb28e7dedb9da39f495ab7af0


.github/actions/setup-e2e-tools/action.yml
```diff
@@ -0,0 +1,13 @@
+name: 'setup-e2e-tools'
+description: 'Set up required tooling for creating an e2e test environment'
+
+runs:
+  using: "composite"
+  steps:
+    - name: 'Install Skaffold'
```
bradegler, line 7: There is a [GitHub action](https://github.com/marketplace/actions/skaffold-github-action) on the marketplace for installing and configuring Skaffold which might make this more flexible in the future. It might be overkill for this use case but I would suggest you take a look at it.

.github/actions/setup-e2e-tools/action.yml
```diff
@@ -0,0 +1,13 @@
+name: 'setup-e2e-tools'
+description: 'Set up required tooling for creating an e2e test environment'
+
```
bradegler, line 3: It might not be necessary now, but exposing some of the important action configuration options as inputs with sane defaults would go a long way to making this more useable. If nothing else, being able to pin to a specific version of `skaffold` or `minikube` would be good.

bradegler (commented):
I'm a little worried about the value of this action. Combining two actions into one action is fine if there is something specific to that combination that we are providing but just putting them together in this case with the default configurations doesn't feel like it does much. Perhaps there is a bigger picture here that I'm missing?

verbanicm:
+1 to Brads comments, can we make this action encompass more of the testing logic instead of just the setup? I am not sure if this makes it not as re-usable then.

.github/actions/setup-e2e-tools/action.yml
sethvargo: I was going to make the same comment. I'm not sure why we're downloading this ourselves, and I'm _very_ worried about always downloading "latest".

sethvargo (commented):
Brad and Mike were online before me, but I also want to +1 their feedback. It feels like it'd be easier to provide copy-paste YAML instead of this composite action (which will still need copy-paste YAML that's not much shorter). Since both installing skaffolding and minikube are pre-packaged in other actions, I'm not sure I understand the overall value.

.github/actions/setup-e2e-tools/action.yml
agarwalnit: Thanks Brad and Seth for your recommendations here. Frankly, I was also slightly skeptical in start but after I did pilot for WISE i realize it is going to be a common pattern where client may need simple wrapper to set up all their required tooling. 

It is two steps now but which may not add much of the value and probably a documentation with some copy / paste it can be solved. But then we anticipate it to grow to 10+ and maintaining cost of version / steps will go up if different client has to do it themselves.

The github marketplace recommendation seems great and we can definitely switch to it.   The underlying minikube action is a good example of defining such reusability which install and start the minikube with some defaults.

agarwalnit:
> +1 to Brads comments, can we make this action encompass more of the testing logic instead of just the setup? I am not sure if this makes it not as re-usable then.

Thanks, yes in the current form it seem like 2 steps and face value provide very little value.   The other testing logic is very much language / repository setup specific , hence I  did not include here.

sethvargo:
I'd like to learn more about the use cases. I suspect individual teams will want to control their versions (instead of being centrally managed)

agarwalnit:
> I'd like to learn more about the use cases. I suspect individual teams will want to control their versions (instead of being centrally managed)

The use case here is use something reusable which help them setup necessary tooling.  The teams I talk to have very little to no opinion and lean towards using something that we give it to them.  On point of teams wanting to control their version, I agree some may but majority probably will not. The teams I talk to prefer an interface over minikube/ skaffold instead of directly being exposed to. 

I can hold  off for now, If you still think we are getting ahead of ourself in providing this reusable action and probably we evaluate set up for few more teams and then provide a reusable action.

sethvargo:
Sure, but I don't think they are that reusable. At the end of the day, we're talking about the customer copy-pasting:

```yaml
- steps: 
  - uses: 'setup-skaffold'
  - uses: 'setup-minikube'
```

vs

```yaml
- steps: 
  - uses: 'setup-minikube-and-skaffold'
```

agarwalnit:
> Sure, but I don't think they are that reusable. At the end of the day, we're talking about the customer copy-pasting:
> 
> ```yaml
> - steps: 
>   - uses: 'setup-skaffold'
>   - uses: 'setup-minikube'
> ```
> 
> vs
> 
> ```yaml
> - steps: 
>   - uses: 'setup-minikube-and-skaffold'
> ```

Makes sense, I will for now discard this PR. We can revisit it once the number of steps increases for our clients or even these two steps start adding friction points.

PR #49, opened 2022-12-23
fix: don't close DB connection after it's established

drevell:
The docs for DB.Close() specfically say not to do this. These connections should be long-lived. This bug may be causing problems with connection pooling, causing clients to get "bad connection" errors.

### DIFF
--- a/mysqltest/docker.go
+++ b/mysqltest/docker.go
@@ -159,7 +159,6 @@ func tryLogin(conf *config, port string) error {
 	if err != nil {
 		return fmt.Errorf("sql.Open(): %w", err)
 	}
-	defer db.Close()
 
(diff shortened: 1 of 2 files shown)


kevya-google:
Can you also fix this in docker_test.go for consistency? https://github.com/abcxyz/pkg/blob/mysqltest_noclose/mysqltest/docker_test.go#L49

kevya-google (approved):
Looks good!

drevell:
> Can you also fix this in docker_test.go for consistency? https://github.com/abcxyz/pkg/blob/mysqltest_noclose/mysqltest/docker_test.go#L49

Thanks for catching that!

PR #50, opened 2023-01-04
Move linter config into a file, add new linters

sethvargo:
This moves the linting configuration into a file and adds a few new linters that were released since the initial design. One of the nicest ones is goheader, which checks that the Apache license header is added to new Go files. Turns out, we had a slight typo that has been copy-pasted around, so I fixed that too.

### DIFF
--- a/.github/workflows/go-lint.yml
+++ b/.github/workflows/go-lint.yml
@@ -7,6 +7,10 @@ on:
         description: 'The version of Go to install and use.'
         type: 'string'
         required: true
+      golangci_url:
+        description: 'The URL to a golangci file. This is only used if no file is found in the local directory.'
+        type: 'string'
+        default: 'https://raw.githubusercontent.com/abcxyz/pkg/main/.golangci.yml'
       directory:
         description: 'Directory in which Go files reside.'
         type: 'string'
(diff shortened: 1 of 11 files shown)


.github/workflows/go-lint.yml
```diff
@@ -61,56 +61,15 @@ jobs:
...
-              "stylecheck",
-              "tenv",
-              "thelper",
-              "tparallel",
-              "typecheck",
-              "unconvert",
-              "unused",
-              "whitespace",
-            ]'), ',') }} --max-issues-per-linter=0 --max-same-issues=0 --timeout=5m
+            --config "${GITHUB_ACTION_PATH}/.golangci.yml"
```
sethvargo, line 68: It doesn't look like `${{ github.acton_path }}` or `${GITHUB_ACTION_PATH}` is set...

testutil/grpc.go
```diff
@@ -32,6 +32,7 @@ type RegisterFunc func(*grpc.Server)
 // establish.
 func FakeGRPCServer(tb testing.TB, registerFunc RegisterFunc) (string, *grpc.ClientConn) {
 	tb.Helper()
+	_ = 1
```
yolocs, line 35: Wonder what's the purpose of this line...

yolocs (approved):

testutil/grpc.go
sethvargo: I was testing something. I'm trying to figure out why sometimes the linter ignores files. I think it's only looking at files modified in the PR.

.golangci.yml
```diff
@@ -0,0 +1,128 @@
...
+        YEAR_AUTHOR: '\d{4} .*'
+    # default: ""
+    template: |-
+      Copyright {{ YEAR_AUTHOR }}
+
+      Licensed under the Apache License, Version 2.0 (the "License");
+      you may not use this file except in compliance with the License.
+      You may obtain a copy of the License at
+
+          http://www.apache.org/licenses/LICENSE-2.0
```
verbanicm, line 118: When using the [addlicense](https://github.com/google/addlicense) tool i suggested, there are 5 spaces before the apaches license url. I think we should find a tool that uses 4 or just make this 5 spaces and standardize on it.

sethvargo: Still a WIP, but I made it regex to match 2+ spaces/tabs

.github/workflows/go-lint.yml
```diff
@@ -60,57 +64,36 @@ jobs:
...
 
+      - name: 'Lint (download default configuration)'
+        id: 'load-default-config'
+        if: ${{ hashFiles('.golangci.yml', '.golangci.yaml') == '' }}
+        run: |-
+          # Create a unique output file outside of the checkout.
+          OUTPUT_FILE="${RUNNER_TEMP}/${GITHUB_SHA:0:7}.golangci.yml"
+
+          # Download the file, passing in authentication to get a higher rate
+          # limit.
```
yolocs, line 75: Interesting. Can we add a reference link here?

.github/workflows/go-lint.yml
```diff
@@ -60,57 +64,36 @@ jobs:
...
+          curl "${{ inputs.golangci_url }}" \
+            --silent \
+            --fail \
+            --location \
+            --header "Authorization: Token ${{ github.token }}" \
+            --output "${GOLANGCI_YML}"
+
+          # Save the result to an output.
+          echo "::notice::Wrote configuration file to ${GOLANGCI_YML}"
+          echo "{output-file}=${OUTPUT_FILE}" >> "${GITHUB_OUTPUT}"
```
yolocs, line 85: ?
```suggestion
          echo "{output-file}=${GOLANGCI_YML}" >> "${GITHUB_OUTPUT}"
```

sethvargo:
@yolocs see https://github.com/abcxyz/lumberjack/pull/362 for what kind of changes we'd be looking at.

Do we want to enable gci to enforce import ordering? https://github.com/daixiang0/gci. It'll make things a bit more complicated, but consistent.

yolocs:
The change in https://github.com/abcxyz/lumberjack/pull/362 looks good to me. 
The consistency from https://github.com/daixiang0/gci looks neat. It would be great if we can somehow directly use it in editor as a post-save hook.

sethvargo:
> It would be great if we can somehow directly use it in editor as a post-save hook.

It has no "package awareness", so we'd have to pipe the package name in manually. That's my main hesitation.

yolocs:
> It has no "package awareness", so we'd have to pipe the package name in manually. That's my main hesitation.

Yea, that might be too much a hassle. After all, manually grouping the imports might just work if we follow the same convention.

verbanicm:
LGTM

PR #52, opened 2023-01-06
feat: add a util to check if it's in an integration test

yolocs:
Use this to skip integration test in unit test so we can run test uniformly with `go test ./...`

```go
func TestMyIntegration(t *testing.T) {
  testutil.SkipIfNotIntegration()
  // Test stuff
}
```

Follow up - update lumberjack / jvs to use this to skip integration test.

### DIFF
--- a/testutil/integ.go
+++ b/testutil/integ.go
@@ -18,6 +18,7 @@ import (
 	"log"
 	"os"
 	"strconv"
+	"testing"
 )
 
 // IsIntegration checks env var TEST_INTEGRATION and consider that we're in an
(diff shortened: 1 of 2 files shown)


testutil/integ.go
```diff
@@ -0,0 +1,45 @@
...
+// IsIntegration checks env var TEST_INTEGRATION and consider that we're in an
+// integration test if it's set to true.
+func IsIntegration() bool {
+	integVal := os.Getenv("TEST_INTEGRATION")
+	if integVal == "" {
+		return false
+	}
+	isInteg, err := strconv.ParseBool(integVal)
+	if err != nil {
+		log.Printf("WARNING: failed to parse TEST_INTEGRATION; will default to 'false': %v", err)
```
drevell, line 33: Optional: you might consider returning an error here. If the user provides TEST_INTEGRATION=TRUe (with an erroneous lower-case "e") then the test will be silently skipped, and the user could be misled into thinking that their test passed.

This might be overengineering 😄

drevell (approved):

testutil/integ.go
yolocs: I've debated with myself :) But now you voted...

drevell (approved):
LGTM

PR #56, opened 2023-01-24
feat: added Proto conversion util

capri-xiyue:
added Proto conversion util


protoutil/conversion_test.go
```diff
@@ -0,0 +1,72 @@
...
+	t.Parallel()
+
+	cases := []struct {
+		name          string
+		input         any
+		want          *structpb.Struct
+		wantErrSubstr string
+	}{
+		{
+			name: "success",
```
yolocs, line 37: To have a failure case, consider: https://stackoverflow.com/questions/33903552/what-input-will-cause-golangs-json-marshal-to-return-an-error

capri-xiyue: done

protoutil/conversion.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+// ToProtoStruct converts v, which must marshal into a JSON object, into a proto struct.
+func ToProtoStruct(v any) (*structpb.Struct, error) {
```
sethvargo, line 26: @yolocs should we genericify this? 

```go
func ToProto[P proto.Message](v any) (P, error)
```

That might not actually work, but it would give us a generic function to convert any thing to a specific proto type.

sethvargo:
```go
func ToProto[T proto.Message](v any) (T, error) {
	var t T

	jb, err := json.Marshal(v)
	if err != nil {
		return t, fmt.Errorf("json.Marshal: %w", err)
	}
	if err := json.Unmarshal(jb, &t); err != nil {
		return t, fmt.Errorf("failed to unmarshal json: %w", err)
	}
	return t, nil
}
```

sethvargo: func ToProto[T proto.Message](v any) (T, error) {
	var t T

	jb, err := json.Marshal(v)
	if err != nil {
		return t, fmt.Errorf("json.Marshal: %w", err)
	}
	if err := json.Unmarshal(jb, &t); err != nil {
		return t, fmt.Errorf("failed to unmarshal json: %w", err)
	}
	return t, nil
}

yolocs: First I don't think this code would compile :) because proto.Message is an interface (remember our earlier discussion?)

I don't see much value to make this generic. Proto struct works because you can unmarshal pretty much any json object into it. But that's not true for other proto types.

sethvargo (approved):

PR #62, opened 2023-02-22
Add an iaputil package

shan4993:
Edit: add new method to existing grpcutil, dont create a new util package.

Im proposing adding an `iaputil` package in order to distinguish the current util which expects JWTs to be provided via GRPC metadata. For context I want to leverage this package to create a [processor like this](https://github.com/abcxyz/jvs/blob/main/pkg/justification/processor.go#L56-L64) but more like `NewIAPProcessor`. This processor would be used by the Controller that renders the UI. I can go into more detail about this but putting out the PR for the JVS code that would use this new util would be more useful. 

Either way looking to get feedback on the change to this repo first.


grpcutil/grpc_jwt_auth.go
```diff
@@ -26,8 +26,8 @@ import (
 	grpcmetadata "google.golang.org/grpc/metadata"
 )
 
-// JWTAuthenticationHandler allows for retrieving principal information from JWT tokens stored in GRPC metadata.
-type JWTAuthenticationHandler struct {
+// GrpcJWTAuthenticationHandler allows for retrieving principal information from JWT tokens stored in GRPC metadata.
```
shan4993, line 29: Maybe this should be `GRPCJWTAuthenticaionHandler`

iaputil/iap_jwt_auth.go
```diff
@@ -0,0 +1,79 @@
...
+	Endpoint string
+}
+
+// NewIAPJWTAuthenticationHandler returns a IAPJWTAuthenticationHandler with a verifier initialized. Uses defaults
+// for JWT related fields that will retrieve a user email when using IAP on GCP.
+func NewIAPJWTAuthenticationHandler(ctx context.Context) (*IAPJWTAuthenticationHandler, error) {
+	j := &IAPJWTAuthenticationHandler{
+		PrincipalClaimKey: "email",
+		// For context: https://cloud.google.com/iap/docs/signed-headers-howto#verifying_the_jwt_header
+		Endpoint: "https://www.gstatic.com/iap/verify/public_key-jwk",
```
shan4993, line 40: The IAP docs have their own [jwt validator](https://github.com/googleapis/google-api-go-client/blob/b35900aff148bb8446ca7e15ed0beb50d749d6c7/idtoken/validate.go#L125-L165) which also validates the aud if/when we choose to validate that. 

If it doesn't matter I continued to leverage the one we are using (github.com/lestrrat-go/jwx/v2/jwt) for the IAP case. Idk if we need to find the public key associated to the kid of the JWT

From the IAP docs: 
https://www.gstatic.com/iap/verify/public_key. This URL contains a JSON dictionary that maps the kid claims to the public key values.
https://www.gstatic.com/iap/verify/public_key-jwk. This URL contains the IAP public keys in [JWK](https://tools.ietf.org/html/rfc7517) format.

sethvargo:
What problem are you trying to solve? We should be able to validate IAP JWTs using the existing verification mechanisms without another library.

Furthermore, if the processor is specific to lumberjack, we should probably keep it in lumberjack instead of having it in pkg.

shan4993:
> What problem are you trying to solve? We should be able to validate IAP JWTs using the existing verification mechanisms without another library.
> 
> Furthermore, if the processor is specific to lumberjack, we should probably keep it in lumberjack instead of having it in pkg.

From the JVS UI ([at this line](https://github.com/abcxyz/jvs/blob/main/pkg/controller/controller.go#L146)), I want to send a jwt string to the createToken api. That API today assumes the the JWT is in the grpc request. 

Additionally the current JWTAuthenticationHandler creator hard codes a [JWTPrefix and JWTKey](https://github.com/abcxyz/pkg/blob/main/grpcutil/jwt_auth.go#L49-L54). These dont apply to the IAP flow. By creating a separate struct here I was attempting to avoid leaving these two fields uninitialized. Maybe there is a better way than my approach.

sethvargo:
@shan4993 do we still need this?

shan4993:
Nope closing! Thanks again.

PR #68, opened 2023-03-02
add github token for terraform-lint

verbanicm:


shan4993 (approved):

.github/workflows/terraform-lint.yml
```diff
@@ -11,6 +11,10 @@ on:
         description: 'The directory upon which to lint Terraform configurations.'
         type: 'string'
         required: true
+    secrets:
+      gh_token:
```
sethvargo, line 15: Can we name this github_token please :)

sethvargo (changes requested):

.github/workflows/terraform-lint.yml
verbanicm: Ah i forgot i tried this before.....github wont let you:

https://github.com/abcxyz/pkg/actions/runs/4314396435

```
secret name `github_token` within `workflow_call` can not be used since it would collide with system reserved name
```

sethvargo (approved):

PR #70, opened 2023-03-02
secrets cant be using in if syntax

verbanicm:


sethvargo (approved):
Add many docs about why, esp linking to https://github.com/actions/runner/issues/520.

bradegler (approved):

PR #73, opened 2023-03-03
Add general purpose HTML and JSON rendering package

sethvargo:
This brings over a bunch of the high-performance and highly-fault-tolerant HTML and JSON rendering from Exposure Notifications, but decouples it from the EN internals. Of note:

- It does not have a dependency on a specific logger and instead allows providing an `onError` handler; there are no dependencies besides the stdlib.
- Reasonable set of built-in functions with the ability to append custom ones.

The package-level comment provides more details and there are some examples of how the package can be used in renderer_test.go.

### DIFF
--- a/renderer/assets.go
+++ b/renderer/assets.go
@@ -1,0 +1,126 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package renderer
+
+import (
+	"bytes"
+	"crypto/sha512"
+	"encoding/base64"
+	"fmt"
+	htmltemplate "html/template"
+	"io"
+	"io/fs"
+	"path/filepath"
+	"strings"
+	"sync"
+	texttemplate "text/template"
+)
+
+const sriPrefix = "sha512-"
+
+var (
+	cssIncludeTmpl = texttemplate.Must(texttemplate.New(`cssIncludeTmpl`).Parse(strings.TrimSpace(`
+{{ range . -}}
+<link rel="stylesheet" href="/{{.Path}}" integrity="{{.SRI}}" crossorigin="anonymous" referrerpolicy="no-referrer" />
+{{ end }}
+`)))
+
+	cssIncludeTagCache htmltemplate.HTML
+)
+
+var (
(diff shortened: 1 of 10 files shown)


renderer/assets.go
```diff
@@ -0,0 +1,126 @@
...
+				Path: pth,
+				SRI:  integrity,
+			})
+		}
+
+		var b bytes.Buffer
+		if err := tmpl.Execute(&b, list); err != nil {
+			return "", fmt.Errorf("failed to render %s asset: %w", search, err)
+		}
+		result := htmltemplate.HTML(b.String())
```
shan4993, line 106: I know this one! lol

//nolint:gosec // No user-supplied input

shan4993 (approved):

shan4993:
Thanks for doing this 🙌

sethvargo:
> Thanks for doing this 🙌

Yup - I'll write it up to the JVS once I get reviews from other folks. Since it's pretty big, I requested some extra eyes 😄.

renderer/renderer.go
```diff
@@ -0,0 +1,238 @@
...
+
+// Renderer is responsible for rendering various content and templates like HTML
+// and JSON responses. This implementation caches templates and uses a pool of
+// buffers.
+type Renderer struct {
+	// rendererPool is a pool of *bytes.Buffer, used as a rendering buffer to
+	// prevent partial responses being sent to clients.
+	rendererPool *sync.Pool
+
+	// templates is the actually collection of templates. templatesLoader is a
```
verbanicm, line 74:
```suggestion
	// templates is the actual collection of templates. templatesLoader is a
```

verbanicm: Also, `templatesLoader` doesn't exist?

renderer/renderer.go
```diff
@@ -0,0 +1,238 @@
...
+// fails after a partial response has been sent.
+func WithOnError(fn func(err error)) Option {
+	return func(r *Renderer) *Renderer {
+		r.onError = fn
+		return r
+	}
+}
+
+// WithTemplateFuncs registers additional template functions. The renderer
+// includes many helpful functions, but some applications may which to
```
verbanicm, line 118:
```suggestion
// includes many helpful functions, but some applications may wish to
```

verbanicm (approved):

renderer/renderer.go
```diff
@@ -0,0 +1,238 @@
...
+// filesystem on disk. However, it accepts the FS interface for testing and
+// [embed] purposes. Because embed does not perform hot reloading, you may want
+// to use a different fs for development versus production:
+//
+//	//go:embed assets assets/**/*
+//	var _assetsFS embed.FS
+//
+//	func AssetsFS() fs.FS {
+//	  // In dev, just read directly from disk
+//	  if v, _ := strconv.ParseBool(os.Getenv("DEV_NODE")); v {
```
verbanicm, line 29:
```suggestion
//	  if v, _ := strconv.ParseBool(os.Getenv("DEV_MODE")); v {
```

pdewilde:
How does this compare to https://pkg.go.dev/html/template ?

sethvargo:
@pdewilde it uses that package internally.

PR #77, opened 2023-03-07
feat: terraform linter that validates against the abcxyz style guide

bradegler:


internal/tools/terraformlinter/terraform_linter.go
```diff
@@ -0,0 +1,331 @@
...
+	attrForEach                = "for_each"
+	attrCount                  = "count"
+	attrProvider               = "provider"
+	attrSource                 = "source"
+	attrProviderProject        = "project"
+	attrProviderProjectID      = "project_id"
+	attrProviderFolder         = "folder"
+	attrProviderFolderID       = "foler_id"
+	attrProviderOrganization   = "organization"
+	attrProviderOrganizationID = "organization_id"
```
verbanicm, line 68: I believe `org_id` is a valid keyword too

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_organization_iam

bradegler: Added it.

internal/tools/terraformlinter/terraform_linter.go
```diff
@@ -0,0 +1,335 @@
...
+	trailingNewline bool
+}
+
+// keywords to match on.
+const (
+	attrForEach                = "for_each"
+	attrCount                  = "count"
+	attrProvider               = "provider"
+	attrSource                 = "source"
+	attrProviderProject        = "project"
```
verbanicm, line 63: Can we enforce the ordering:

```
org
folder
project
```

verbanicm: Can be a follow up as I assume thats not a 2 line fix.

bradegler: Your doc doesn't say it needs to be ordered last I checked. I was going to order them but then thought "woohoo I can skip that part!"

I'll get it in a follow up unless I get a bunch of other changes I need to do anyhow.

verbanicm: Updated doc. Thanks.

verbanicm:
Is there anyway to ensure all terraform resources are created with underscore separated names?

eg `my_test_project`

This would include, resources, modules, variables, outputs I believe (am I missing anything)?

See the Other Naming section in the doc.

bradegler:
> 

That shouldn't be too hard.

Look for `-` instead of `_`? Any other bad characters?

internal/tools/terraformlinter/terraform_linter.go
```diff
@@ -0,0 +1,335 @@
...
+	// Process each provided path looking for violations
+	for _, path := range paths {
+		instances, err := lint(path)
+		if err != nil {
+			return fmt.Errorf("error linting files: %w", err)
+		}
+		violations = append(violations, instances...)
+	}
+	for _, instance := range violations {
+		fmt.Printf("%q detected at [%s:%d]\n", instance.ViolationType, instance.Path, instance.Line)
```
sethvargo, line 119: Should this be more structured so we can use `reviewdog` or a similar tool to do GitHub-level annotations? I don't think the linter should write to GitHub Actions directly.

internal/tools/terraformlinter/terraform_linter.go
```diff
@@ -0,0 +1,335 @@
...
+		return fmt.Errorf("found %d violation(s)", len(violations))
+	}
+
+	return nil
+}
+
+// lint reads a path and determines if it is a file or a directory.
+// When it finds a file it reads it and checks it for violations.
+// When it finds a directory it calls itself recursively.
+func lint(path string) ([]*ViolationInstance, error) {
```
sethvargo, line 131: Not required change, just an FYI. If you accept an `fs.FS`, it makes testing easier because you can create an in-memory FS pretty easily.

internal/tools/terraformlinter/terraform_linter.go
```diff
@@ -0,0 +1,335 @@
...
+	isDir, err := isDirectory(path)
```
sethvargo, line 132: I think you've unintentionally re-implemented `filepath.Walk`. It might clean things up here.

internal/tools/terraformlinter/terraform_linter.go
```diff
@@ -0,0 +1,335 @@
...
+				instances = append(instances, &ViolationInstance{ViolationType: fmt.Sprintf(errorTrailingMetaBlockAttribute, attrDependsOn), Path: token.token.Range.Filename, Line: token.token.Range.Start.Line})
+			}
+		case attrLifecycle:
+			// lifecycle should be last
+			if pos != len(idents)-1 {
+				instances = append(instances, &ViolationInstance{ViolationType: fmt.Sprintf(errorTrailingMetaBlockAttribute, attrLifecycle), Path: token.token.Range.Filename, Line: token.token.Range.Start.Line})
+			}
+		// All provider specific entries follow the same logic. Should be below the metadata segment and above everything else
+		case attrProviderProject:
+			fallthrough
```
sethvargo, line 303: I think this is the first time I've seen `fallthrough` used. Why do we need it?

internal/tools/terraformlinter/terraform_linter_test.go
```diff
@@ -0,0 +1,662 @@
...
+
+	cases := []struct {
+		name      string
+		content   string
+		filename  string
+		expect    []*ViolationInstance
+		wantError bool
+	}{
+		{
+			name:      "No special attributes",
```
sethvargo, line 282: Oops, this diff was suppressed by GitHub.

1. `name` attributes should prefer lowercase and have no spaces (use underscores). It helps with grepping for failed tests.

1. I think we should inline the `content` so the test is fully self-contained.

internal/tools/terraformlinter/terraform_linter.go
```diff
@@ -0,0 +1,335 @@
...
+	attrProviderFolderID:       Provider,
+	attrProviderOrganization:   Provider,
+	attrProviderOrganizationID: Provider,
+	attrProviderOrgID:          Provider,
+	attrDependsOn:              Trailing,
+	attrLifecycle:              Trailing,
+}
+
+const (
+	errorLeadingMetaBlockAttribute  = "The attribute %q must be in the meta block at the top of the definition."
```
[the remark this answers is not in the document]
bradegler: These aren't really errors that are throw, just the description of what the linter saw as a violation. I'll call them something other than error to reduce confusion.

internal/tools/terraformlinter/terraform_linter.go
bradegler: Switched to using filepath.WalkDir.

internal/tools/terraformlinter/terraform_linter.go
bradegler: Mainly because I hate repeating myself but I'm going to have to change this to force ordering on these attributes anyhow so there won't be any opportunity to fallthrough after that change.

internal/tools/terraformlinter/terraform_linter.go
bradegler: I'll keep that in mind for the future. Right now its a string on the command line and filepath.WalkDir wants a string so converting seems wasteful.

internal/tools/terraformlinter/terraform_linter.go
```diff
@@ -0,0 +1,315 @@
...
+		case attrProviderProjectID:
+			fallthrough
+		case attrProviderFolder:
+			fallthrough
+		case attrProviderFolderID:
+			fallthrough
+		case attrProviderOrganization:
+			fallthrough
+		case attrProviderOrganizationID:
+			fallthrough
```
sethvargo, line 293: You can condense this a bit (not required):

```suggestion
		case attrProviderProject,
		  attrProviderProjectID,
		  attrProviderFolder,
		  attrProviderFolderID,
		  attrProviderOrganization,
		  attrProviderOrganizationID:
			fallthrough
```

sethvargo (approved):

internal/tools/terraformlinter/terraform_linter.go
bradegler: No need for the fallthrough at all, just matching on all of the possible values. I didn't realize I could do that and makes me question what the use case for fallthrough actually is since this is much cleaner.

internal/tools/terraformlinter/terraform_linter.go
bradegler: Filed https://github.com/abcxyz/pkg/issues/78 and will fix this in there.

PR #79, opened 2023-03-08
Deprecate jwtutil

sethvargo:
Now that we've standardized on a single library, it doesn't provide much value beyond using the library directly. We also have some unique use cases that require exposing lower-level primitives in lestrrat-go/jwx too.


shan4993 (approved):
I know for IAP there is a [go library](https://github.com/googleapis/google-api-go-client/blob/v0.111.0/idtoken/validate.go#L112) that does further validation of the aud if its provided. When we switch over to verifying the JWT will this current library be enough to extract the necessary information (in this case the user's email)? 

I suppose the intention is to stick to a single library for JWT related functionality if possible.

jwtutil/jwt.go
```diff
@@ -1,62 +0,0 @@
...
-	"github.com/lestrrat-go/jwx/v2/jwk"
-	"github.com/lestrrat-go/jwx/v2/jws"
-	"github.com/lestrrat-go/jwx/v2/jwt"
-)
-
-const IAMKeyEndpoint = "https://www.googleapis.com/oauth2/v3/certs"
-
-// Verifier allows for getting public JWK keys from an endpoint and validating JWTs with
-// those keys.
-type Verifier struct {
```
yolocs, line 31: Why removing this? Is the idea to use lestrrat-go/jwx directly to validate token?

This is still used in some minty I think: https://cs.github.com/?scope=org%3Aabcxyz&scopeName=abcxyz&q=NewVerifier

sethvargo: Yup. It's not used in Minty anymore :). Minty needed access to lower parts of the library, which motivated this change.

yolocs (approved):

PR #88, opened 2023-03-16
fix: fixed unmarshal yaml bug

capri-xiyue:
fixed unmarshal yaml bug. Previously it can't handle map

### DIFF
--- a/protoutil/conversion.go
+++ b/protoutil/conversion.go
@@ -39,8 +39,8 @@ func ToProtoStruct(v any) (*structpb.Struct, error) {
 	return x, nil
 }
 
-// UnmarshalYAML unmarshals the give YAML bytes to the given proto message.
-func UnmarshalYAML(b []byte, msg proto.Message) error {
+// FromYAML converts YAML to Proto.
+func FromYAML(b []byte, msg proto.Message) error {
 	tmp := map[string]any{}
 	if err := yaml.Unmarshal(b, tmp); err != nil {
 		return fmt.Errorf("failed to unmarshal yaml: %w", err)
--- a/protoutil/conversion.go
+++ b/protoutil/conversion.go
@@ -22,7 +22,8 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/structpb"
-	"gopkg.in/yaml.v2"
+
+	ghodssyaml "github.com/ghodss/yaml"
 )
 
 // ToProtoStruct converts v, which must marshal into a JSON object, into a proto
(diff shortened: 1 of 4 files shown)


protoutil/conversion.go
```diff
@@ -39,18 +40,13 @@ func ToProtoStruct(v any) (*structpb.Struct, error) {
...
 
-// UnmarshalYAML unmarshals the give YAML bytes to the given proto message.
-func UnmarshalYAML(b []byte, msg proto.Message) error {
-	tmp := map[string]any{}
-	if err := yaml.Unmarshal(b, tmp); err != nil {
-		return fmt.Errorf("failed to unmarshal yaml: %w", err)
-	}
-	jb, err := json.Marshal(tmp)
+// YAMLToProto converts YAML to Proto.
+func YAMLToProto(y []byte, m proto.Message) error {
```
capri-xiyue, line 44: I feel renaming makes the functionality more clear. I can search the existing usage of "UnmarshalYAML" and change it to "YAMLToProto". Let me know if there is any concerns.

sqin2019 (approved):

protoutil/conversion.go
```diff
@@ -22,7 +22,8 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/structpb"
-	"gopkg.in/yaml.v2"
+
+	ghodssyaml "github.com/ghodss/yaml"
```
sethvargo, line 26: Is there a reason we're switching packages? `gopkg.in/yaml` is the most popular one. I'm not really supportive of bringing in a bespoke YAML library.

protoutil/conversion.go
pdewilde: I think it makes sense. UnmarshalYAML makes me think its more general purpose.

protoutil/conversion.go
pdewilde: Seems like its sitting on top of gopkg.in/yaml and is good if you need to support both yaml and json, see https://web.archive.org/web/20150812020634/http://ghodss.com/2014/the-right-way-to-handle-yaml-in-golang/

capri-xiyue: gopkgin.yaml doesn't work here, it can't handle maps.  gopkgin.yaml is also used by k8s sig https://pkg.go.dev/sigs.k8s.io/yaml#section-readme

protoutil/conversion.go
sethvargo, line 44: This is now `protoutil.YAMLToProto`. I wonder if `protoutil.FromYAML` might be easier?

protoutil/conversion.go
yolocs: Yes, `gopkg.in/yaml` unmarshals yaml nests into `map[interface{}]interface{}` rather than `map[string]interface{}`, which in turn causes problem with json marshaling.

yolocs: For the record, I've used `github.com/ghodss/yaml` before and know it's being used in some other well-known OSS projects. I think it's easier to use than `gopkg.in/yaml`

protoutil/conversion.go
yolocs: The package name is `protoutil`. `protoutil.UnmarshalYAML` seems clear enough?

protoutil/conversion.go
sethvargo: I'm not sure I understand. It works fine with the gopkg.in/yaml.v3: https://gist.github.com/sethvargo/b11a0e22886edcc3c61e95a5a375f588

sethvargo (changes requested):

protoutil/conversion.go
yolocs: Yea, pretty nit but in general I'd like to avoid duplicated info that already presents in the package name.

protoutil/conversion.go
yolocs: I was about to ask - have we tried v3 yet?

capri-xiyue: done

protoutil/conversion.go
capri-xiyue: done

capri-xiyue: As Seth suggested, I changed it to 'FromYAML', is that ok?

sethvargo (approved):

yolocs (approved):

PR #101, opened 2023-03-27
Add IsIntegrationMain

yolocs:
To check if it's in integration test when `testing.TB` is not accessible, e.g. in `TestMain`

### DIFF
--- a/testutil/integ.go
+++ b/testutil/integ.go
@@ -15,6 +15,7 @@
 package testutil
 
 import (
+	"log"
 	"os"
 	"strconv"
 	"testing"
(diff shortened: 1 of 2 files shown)


testutil/integ.go
```diff
@@ -15,6 +15,7 @@
 package testutil
 
 import (
+	"log"
```
spencergoldberg, line 18: Why not our internal logger?

yolocs: You mean https://github.com/abcxyz/pkg/tree/main/logging?

No need. For test output, we don't have to use structured logging. We expect an exit right away.

sethvargo: Yea - our logger is for structured application level logging. That being said, it might be easier to just `return false` instead of logging a message.

sethvargo (commented):
This is fine, but I'm not sure what value it provides over just doing this inline in a `TestMain`? I worry about the indirection of the envvar being buried further down in a package.

testutil/integ.go
yolocs: I vaguely remember a quick discussion about this - if we return `false` when the env var value is not a boolean, a typo (e.g. `truu`) could "sneakily" skip the test.

yolocs:
> This is fine, but I'm not sure what value it provides over just doing this inline in a `TestMain`? I worry about the indirection of the envvar being buried further down in a package.

I'm probably seeking artificial consistency here - did this mostly because we already have a helper func `IsIntegration` but it won't work for `TestMain`. Happy to drop this.

sethvargo:
"Artificial consistency" :)

I would drop it until we see a major divergence we need to correct.

PR #103, opened 2023-04-03
Add general-purpose serving infrastructure

sethvargo:
This standardizes our http/grcp server creation and lifecycle handling.

### DIFF
--- a/serving/serving.go
+++ b/serving/serving.go
@@ -1,0 +1,201 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package serving provides an extremely opinionated serving infrastructure,
+// with support [net/http.Server] and [google.golang.org/grpc.Server].
+//
+// It supports listening on specific ports or randomly-available ports, and
+// reports the corresponding bind addresses. This is useful for creating
+// multiple servers in tests where ports may conflict. The server also
+// gracefully stops requests when the provided context is closed.
+package serving
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/abcxyz/pkg/logging"
+	"google.golang.org/grpc"
+)
+
+// Server provides a gracefully-stoppable http server implementation.
+type Server struct {
+	ip       string
+	port     string
+	listener net.Listener
+}
+
+// New creates a new HTTP server listening on the provided address that responds
+// to the http.Handler. It starts the listener, but does not start the server.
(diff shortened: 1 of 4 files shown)


serving/serving.go
```diff
@@ -0,0 +1,201 @@
...
+		if err := srv.Serve(s.listener); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			select {
+			case errCh <- err:
+			default:
+			}
+		}
+	}()
+
+	// Wait for the provided context to finish.
+	<-ctx.Done()
```
yolocs, line 123: If the context is not done but `Serve` func returns an error, would this unintentionally block the code? Do we even want to block if there is an error?

serving/serving.go
```diff
@@ -0,0 +1,201 @@
...
+		if err := srv.Serve(s.listener); err != nil && !errors.Is(err, grpc.ErrServerStopped) {
+			select {
+			case errCh <- err:
+			default:
+			}
+		}
+	}()
+
+	// Wait for the provided context to finish.
+	<-ctx.Done()
```
yolocs, line 190: Same question

serving/serving.go
sethvargo: Oh yea, that would be a bad bug 😄. Fixed - thanks for catching.

serving/serving.go
```diff
@@ -0,0 +1,201 @@
...
+	logger := logging.FromContext(ctx)
+
+	// Start the server in a background goroutine so we can listen for cancellation
+	// in the main process.
+	errCh := make(chan error, 1)
+	doneCh := make(chan struct{})
+	go func() {
+		defer close(doneCh)
+
+		logger.Infow("server is starting", "ip", s.ip, "port", s.port)
```
pdewilde, line 111: I'm curious, why use the non-sentence convention for log messages? Wouldn't it make sense for these?

yolocs (approved):

pdewilde:
Another question: would it make sense to push this out to sandwich?

serving/serving.go
```diff
@@ -0,0 +1,201 @@
...
+// Note that the incoming [net/http.Server]'s address is ignored over the
+// listener's configuration.
+func (s *Server) StartHTTP(ctx context.Context, srv *http.Server) error {
+	logger := logging.FromContext(ctx)
+
+	// Start the server in a background goroutine so we can listen for cancellation
+	// in the main process.
+	errCh := make(chan error, 1)
+	doneCh := make(chan struct{})
+	go func() {
```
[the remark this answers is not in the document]
sethvargo: Yea - I had that thought and went down that path. It ends up adding a layer of indirection that felt more harmful than helpful.

sethvargo:
> Another question: would it make sense to push this out to sandwich?
> 
> Or is there value there in having less indirection for someone new to the codebase?

It depends on how "production ready" you want the examples in sandwhich to be.

pjh:
> It depends on how "production ready" you want the examples in sandwhich to be.

Can you elaborate? We intend to write production-ready code for Sandwich, though we're getting there iteratively.

It's not clear to me if the interface in this PR is something we'll want to adopt or even be able to adopt in Sandwich (given our use of Chi there), but from the PR description it sounds like we probably should.

sethvargo:
> It's not clear to me if the interface in this PR is something we'll want to adopt or even be able to adopt in Sandwich (given our use of Chi there), but from the PR description it sounds like we probably should.

This is for _serving_, Chi is for _routing_. Chi implements `http.Handler`, so a Chi instance could be passed to `ServeHTTPHandler` or `ServeHTTP` with an `http.Server#Handler`.

PR #105, opened 2023-04-04
create terraform lint input to prevent walking child dirs

verbanicm:
Fixes https://github.com/abcxyz/gcp-org-terraform-template/pull/22 issue


sethvargo:
Hey @verbanicm - I don't fully understand what this is for? The input seems kind of confusing, and I'm not sure why it's needed (even after reading https://github.com/abcxyz/gcp-org-terraform-template/pull/22)

PR #106, opened 2023-04-07
feat: add shortcut for writing stdout and stderr in the cli package

yolocs:


cli/command.go
```diff
@@ -61,10 +61,14 @@ type Command interface {
 	// Stdout returns the stdout stream. SetStdout sets the stdout stream.
 	Stdout() io.Writer
 	SetStdout(w io.Writer)
+	// Outf is a shortcut to write to [Command.Stdout].
```
sethvargo, line 64:
```suggestion

	// Outf is a shortcut to write to [Command.Stdout].
```

cli/command.go
```diff
@@ -61,10 +61,14 @@ type Command interface {
...
+	Outf(format string, a ...any)
 
 	// Stderr returns the stderr stream. SetStderr sets the stderr stream.
 	Stderr() io.Writer
 	SetStderr(w io.Writer)
+	// Errf is a shortcut to write to [Command.Stderr].
```
sethvargo, line 70:
```suggestion

	// Errf is a shortcut to write to [Command.Stderr].
```

cli/command_test.go
```diff
@@ -314,7 +314,7 @@ func (c *TestCommand) Run(ctx context.Context, args []string) error {
 	}
 
 	if v := c.Output; v != "" {
-		fmt.Fprint(c.Stdout(), v)
+		c.Outf(v)
```
sethvargo, line 317: I think there might be a few examples to clean up too?

sethvargo (approved):

sethvargo:
Thanks for doing this!

cli/command_test.go
yolocs: Done

PR #109, opened 2023-04-17
Extract mysql from db test to allow for future modularity

pdewilde:


databasetest/mysqltest.go
```diff
@@ -0,0 +1,78 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package databasetest provides an ephemeral database server for testing database integration. It's
+// designed to be used in code that needs to work inside and outside Google.
+package databasetest
```
sethvargo, line 17: This would still need to be in the `mysqltest` package and folder (or have a stub there) to maintain compatibility, if we care.

sethvargo (commented):
This is fine, but I think it would be more common and idiomatic to have dedicated structs that encapsulate each database driver. For example:

```go
type Driver interface {
  Image() string
  TestConn() func(string) string // not sure exact signature
}

type MySQL struct {}

type PostgreSQL struct {}
```

You would need to plumb in some of the data, but I think it's doable

databasetest/mysqltest.go
pdewilde: > Dave Revell:
> 
> Hopefully we can keep it. If we change it, we need to change their google3 and github code to use the new name. The name is used in in google3 in some places that will be quite difficult to change, like https://source.corp.google.com/piper///depot/google3/third_party/golang/github_com/abcxyz/pkg/v/v0/mysqltest/BUILD?q=mysqltest%20f:third_party%20f:golang&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental
> 
>

databasetest/config.go
```diff
@@ -12,15 +12,21 @@
...
+
+	"github.com/ory/dockertest/v3"
+)
 
 // This file implements the "functional options" pattern.
 
 type config struct {
-	killAfterSec   int // This is in integer seconds because that's what Docker takes.
-	mySQLVersion   string
+	waitForPort    string                                  // port to listen on to see if container has started. "" for no port
```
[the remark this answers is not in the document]
pdewilde: Changed to a list of ports

databasetest/config.go
drevell, line 1: Could you add PR comments on this file that note which parts are new and which parts are copied?

databasetest/databasetest.go
```diff
@@ -0,0 +1,49 @@
...
+
+// ConnInfo specifies how to connect to the created container.
+type ConnInfo struct {
+	Hostname string
+	// PortMapper maps from container port to host port. Do not use after container is closed.
+	PortMapper func(string) string
+}
+
+type Driver interface {
+	ImageRepository() string // Repository for docker image (ex: mysql)
```
drevell, line 29: I'm not sure, but I think the comments need to appear above the field they describe. Otherwise they won't show up properly in godoc.

databasetest/databasetest.go
```diff
@@ -0,0 +1,49 @@
...
+import "io"
+
+// ConnInfo specifies how to connect to the created container.
+type ConnInfo struct {
+	Hostname string
+	// PortMapper maps from container port to host port. Do not use after container is closed.
+	PortMapper func(string) string
+}
+
+type Driver interface {
```
drevell, line 28: Could you add a doc comment on the Driver interface? Just a sentence will do.

databasetest/docker.go
```diff
@@ -187,7 +164,7 @@ func (p *containerCloser) Close() error {
 		err = p.pool.Purge(p.container)
 	})
 	if err != nil {
-		return fmt.Errorf("failed stopping MySQL docker container: %w", err)
+		return fmt.Errorf("failed stopping dabase docker container: %w", err)
```
drevell, line 167: "dabase"

databasetest/databasetest.go
```diff
@@ -0,0 +1,49 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package databasetest provides an ephemeral database server for integration testing.
+// It's designed to be used in code that needs to work inside and outside google.
+package databasetest
```
drevell, line 17: (deep thoughts warning) like, dude, is there anything about this that's actually specific to databases? Or is it just a higher level interface to docker test containers?

drevell (changes requested):

databasetest/docker.go
pdewilde: Its where information about District Attorneys is stored

databasetest/databasetest.go
pdewilde: Its definitely the latter, wasn't sure if we wanted to push that usecase or not

databasetest/docker.go
sethvargo: https://www.youtube.com/watch?v=gP3MuUTmXNk

databasetest/databasetest.go
pdewilde: https://bradleyjkemp.dev/godoc-playground/

From what I can tell playing with that and the `go doc` command, godoc doesn't parse interface comments or struct field comments. It just puts the entire interface/struct in a codeblock comments and all, so its purely a style/readability concern.

databasetest/config.go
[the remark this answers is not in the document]
pdewilde: I'm not sure "0" makes sense as a value in this case, as its the container's internal port numbers. The external ports are already "choose for me"

databasetest/databasetest.go
drevell: Seems like just a matter of time until this use case (non-databases) arises. If we can address it now just by using a different name, maybe it's worth it.

sethvargo: I think the "run a container" bit is generalizable, but the port mapping and "wait for ready" seems pretty specialized to databases.

pdewilde: If we want to change, what would be a good name? Its a wrapper for `dockertest`, so I wouldn't want to use that name. I was thinking maybe `testcontainer` but its a little too close to https://golang.testcontainers.org/

`testdockerdeps`? Not a fan of that either

pdewilde: I think I would rename `Driver` to `Service`

pdewilde: Changes made, let me know what you think

databasetest/config.go
```diff
@@ -12,15 +12,21 @@
...
+	startTester    func(Logger, func(string) string) error // function which should return nil error if container has started properly
```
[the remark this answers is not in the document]
pdewilde: Tried to clear  it up a bit

pdewilde:
(general question)

Godoc doesn't seem to parse method comments in interfaces. Does it really make sense to have to re-comment every method I export? Is it redundant to comment  `Port returns default container port` for both Postgres and MySQL?

databasetest/databasetest.go
drevell: > Wouldn't port mapping and wait for ready also apply to pretty much any IP-based dependency? (Say a message queue simulator, some 3p app, shim for some GCP dependency for local testing)

+1, any server with non-negligible startup time would need this, I think.

drevell:
> (general question)
> 
> Godoc doesn't seem to parse method comments in interfaces. Does it really make sense to have to re-comment every method I export? Is it redundant to comment `Port returns default container port` for both Postgres and MySQL?

I'll often just comment "// Port implements the foo.Driver interface" on every method that implements the interface.

mysqltest/mysqltest.go
```diff
@@ -31,13 +32,21 @@ type ConnInfo struct {
...
 func MustStart(opts ...Option) (ConnInfo, io.Closer) {
 	conf := buildConfig(opts...)
-	ci, closer, err := start(conf)
-	if err != nil {
-		// The Closer must be called even if there's an error, to clean up the docker container that may
-		// exist.
-		_ = closer.Close()
-		panic(err)
+	driver := (&containertest.MySQL{}).WithVersion(conf.mySQLVersion)
+	realOpts := make([]containertest.Option, 0, 2)
```
drevell, line 36: `realOpts` is perhaps a bit confusing as a name, because it's not immediately clear why they're more real than the other options. Maybe `ctOpts` or `translatedOpts` or something?

drevell (approved):
LGTM, assuming there's already been a grep-and-replace for occurrences of "mysql" and "database."

containertest/README.md
pjh, line 1: Please update this file, it appears to be very MySQL-specific still.

pdewilde: Good catch

mysqltest/mysqltest.go
pdewilde: Changed

containertest/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+	mySQLService := (&MySQL{}).WithVersion("8.0")
+    ci, closer = containertest.MustStart(mySQLService) // Start the docker container. Can also pass options.
+    exitCode := m.Run() // Runs unit tests
+
+    // Remove container. If tests panic, this won't run, but there's nothing we can do about that:
+    // https://github.com/golang/go/issues/37206#issuecomment-590441512. In that case, then the
+    // container will eventually time out and be cleaned up. Defer isn't used because
+	// os.Exit() will prevent execution, and is needed for TestMain pattern.
+    closer.Close()
```
sethvargo, line 51: There's a trick to make this easier:

```go
func TestMain(m *testing.M) {
  os.Exit(func() int {
    ci, closer := // ...
    defer closer.Close()

    return m.Run()
  })
}
```

sethvargo, line 51: I don't think it's very idiomatic to return a closer _structure_. Usually a closer is a function (two return values) or dangles on the returned object (one return value).

containertest/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+    os.Exit(exitCode)
+}
+
+func TestFoo(t *testing.T) {
+    // One thing you might want to do is create an SQL driver:
```
sethvargo, line 57: `t.Parallel()` 😄

containertest/README.md
```diff
@@ -0,0 +1,112 @@
...
+	
+	m := mySQLService
+	// Find the port docker exposed for your container
+	mySQLPort := ci.PortMapper(m.Port())
+	uri := fmt.Sprintf("%s:%s@tcp(%s)/%s", m.Username(), m.Password(),
+		net.JoinHostPort(ci.Host, mySQLPort), "")
+    db, err := sql.Open("mysql", uri)
+    if err != nil {
+        t.Fatal(err)
+    }
```
sethvargo, line 67: I think this indentation is off.

containertest/README.md
```diff
@@ -0,0 +1,112 @@
...
+)
+
+var ci containertest.ConnInfo
+var mySQLService *containertest.MySQL
+
+// TestMain runs once at startup to do test initialization common to all tests.
+func TestMain(m *testing.M) {
+    var closer io.Closer
+
+	mySQLService := (&MySQL{}).WithVersion("8.0")
```
sethvargo, line 43: This isn't _super_ common style for Go. Usually it would be:

```go
mySQLService := &MySQL{
  Version: "8.0"
}
```

containertest/config.go
```diff
@@ -0,0 +1,83 @@
...
+	killAfterSec   int // This is in integer seconds because that's what Docker takes.
+	service        Service
+	progressLogger Logger
+}
+
+// Logger allows the caller to optionally provide a custom logger for printing status updates about
+// service startup progress. The default is to use the go "log" package.
+type Logger interface {
+	Printf(fmtStr string, args ...any)
+}
```
sethvargo, line 33: Since this is intended to be used by the `testing` package, I'm not sure this is necessary. I wonder if we should accept `testing.TB` and call `t.Log/t.Logf` instead.

containertest/containertest.go
```diff
@@ -0,0 +1,60 @@
...
+package containertest
+
+import "io"
+
+// ConnInfo specifies how connect to the created container.
+type ConnInfo struct {
+	Host string
+
+	// PortMapper maps from container port to host port. Do not use after container is closed.
+	PortMapper func(string) string
```
sethvargo, line 26: Can you make this another type (or name the values)? It's include exactly what the inputs and outputs would be.

containertest/docker.go
```diff
@@ -118,54 +103,49 @@ func runContainer(conf *config, pool *dockertest.Pool) (*dockertest.Resource, er
...
+		for _, waitPort := range conf.service.StartupPorts() {
+			if waitPort != "" {
+				port := container.GetPort(waitPort)
+				if port == "" {
+					notReadyPorts = append(notReadyPorts, port)
+				}
+			}
+		}
+		if len(notReadyPorts) > 0 {
+			return fmt.Errorf("resource.GetPort() returned empty string for port(s) %+q, container isn't ready yet", notReadyPorts)
```
sethvargo, line 130: I think we want to wrap this in a retryable error?

containertest/mysql.go
```diff
@@ -0,0 +1,99 @@
...
+
+	_ "github.com/go-sql-driver/mysql" // Force mysql service to be included.
+)
+
+const (
+	// It's OK to hardcode the root password because only boilerplate test data is stored. Also,
+	// having a well-known password can help with human inspection for debugging. The value chosen for
+	// the password is arbitrary. It can be changed without breaking anything; it's not hardcoded into
+	// the docker image or anything like that.
+	password = "8mo5lfYKjy6ebTK" //nolint:gosec
```
sethvargo, line 30: This can be a follow-up, but since we're abstracting this now, I'd like to generate a random password instead.

containertest/README.md
pdewilde: I was copying the existing style: https://github.com/abcxyz/pkg/blob/main/mysqltest/config.go

Is there a reason the pattern was used there but shouldn't be used here? Or would you also change that to use public fields if you had the chance?

containertest/README.md
pdewilde: Yeah, it should be fixed now.

containertest/README.md
pdewilde: I was following the existing pattern.

Would it make more sense for closer to be a function field in the `ConnInfo` struct which would have a `Close()` method that calls it?


```go
// ConnInfo specifies how connect to the created container.
type ConnInfo struct {
	Host string

	// PortMapper maps from container port to host port. Do not use after container is closed.
	PortMapper func(string) string
	
	closer *containerCloser
}

func (c *ConnInfo) Close() error {
	return c.closer.Close();
}

```

containertest/config.go
pdewilde: That may be better, though wouldn't it require some breaking changes to the `mysqltest` library?

containertest/mysql.go
pdewilde:
``` go
func generatePassword() string {
	return "a random password"
}
```

Sure, I can do. I'll probably just generate a short alphanumeric. Special characters will mess up my uri formatting because I didn't escape the uri.

pdewilde: Though, there is a counter argument:

It could be nice to be able to put a breakpoint in your test code and attach a db client, which would be more difficult if the password changes around

containertest/docker.go
pdewilde: As in define a new error type called retryable? Is there a point if `dockertest` just swallows it in the retryable case?

containertest/containertest.go
pdewilde: I'll rename for now. I was thinking there may be value in changing it to an interface in case we want to extend it in the future, but I'm not sure what else would be added, and it will be more characters to use.

containertest/README.md
pdewilde: Or possibly
```go
// ConnInfo specifies how connect to the created container.
type ConnInfo struct {
	Host string

	// PortMapper maps from container port to host port. Do not use after container is closed.
	PortMapper func(string) string
	
	closer func() error
}

func (c *ConnInfo) Close() error {
	return c.closer();
}
```

containertest/docker_test.go
```diff
@@ -12,19 +12,22 @@
...
 	"fmt"
+	"net"
 	"strings"
 	"testing"
 	"time"
 )
 
+// MySQL will be used as the default service for testing general behavior. Other
+// services should have their own tests written.
 func TestKillAfter(t *testing.T) {
```
pdewilde, line 31: This is flaky on my box :(

containertest/README.md
sethvargo: The internal implementation isn't super important, but it would be most common to have this dangle off the ConnInfo structure as a function like you have in both examples.

containertest/README.md
sethvargo: It's the "options pattern" versus the "struct pattern", but usually the options pattern looks like this:

```go
type Option(a) (a)
func NewThing(ctx context.Context, opts ...Option)
```

Idiomatically, I would expect either:

```go
svc := NewMySQL(ctx, WithVersion("8.0"))
```

or

```go
svc := &MySQL{
  Version: "8.0",
}
```

The "builder pattern" (`foo.WithBar().WithBaz()`) is rarer and usually exists to maintain backwards compatibility with an existing API.

containertest/config.go
sethvargo: What breaking changes do you envision? We can keep the same API (even if we don't use the given logger)

containertest/docker.go
sethvargo: Sorry - I didn't realize we were using a different retry library. In [go-retry](https://github.com/sethvargo/go-retry), you have to explicitly wrap an error as retryable. I guess `pool.Retry` just retries on all errors?

containertest/mysql.go
sethvargo: I don't know if having a function return a random password works, since you'd want it to be deterministic for each instance. Therefore I think you'd need a password field on the structure and then on creation you'd generate a random password. You can avoid the escaping need by cheating a little bit and reading ~16 random bytes and then hex encoding it (which gives you 32 characters).

I agree on the counterargument. My counter-counter arguments are:

- We could print out the password (still kinda sucks)
- You could configure container auth to allow all local network traffic (and therefore no password is required)

containertest/docker.go
pdewilde: Yeah, it doesn't make any distinction

containertest/config.go
pdewilde: Upon further inspection, it appears that WithLogger() is never used, @drevell gave his blessing to remove it, so I will do so.

containertest/README.md
pdewilde: Switched to the struct pattern

containertest/config.go
pdewilde: Does it make sense to require a testing.TB when calling MustStart?

containertest/mysql.go
pdewilde: I think I'll leave as-is for now as none of the proposed solutions would require an interface change. Let me know if you want me to add an issue.

containertest/docker_test.go
pdewilde: I extended startup time to 2 minutes but still sometimes it doesn't start up in time.

pdewilde: I fixed it by forcing 5.7, it seems the latest image (docker 8) is taking a lot longer to boot sometimes.

containertest/config.go
sethvargo: I don't think so, because `MustStart` is intended to be called in `TestMain`, right?

containertest/docker.go
```diff
@@ -0,0 +1,175 @@
...
+// (For databases): Since the startup time for database containers can be as long as 20 seconds,
+// we share the container among every test. Each test should use a randomly-created
+// database/schema name to avoid collisions between tests.
+//
+// This function installs a signal handler for SIGTERM, SIGKILL, and SIGINT that attempts to clean
+// up the Docker container, then runs os.Exit(1). Since the signal handler kills the process, any
+// other custom signal handlers that are installed may not get a chance to run.
+//
+// Docker must be installed on localhost for this to work. No environment vars are needed.
+func start(conf *config) (ConnInfo, error) {
```
sethvargo, line 52: This should return a pointer to a ConnInfo (same with startContainer)

containertest/containertest.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// StartupPorts is the list of ports that must be exposed by container before TestConn is run.
+	StartupPorts() []string
+
+	// TestConn takes a logger and a struct with connection info, and returns nil if app has started.
+	TestConn(progressLogger Logger, info ConnInfo) error
+}
+
+// MustStart starts a container, or panics if there was an error.
+func MustStart(service Service, opts ...Option) ConnInfo {
```
sethvargo, line 58: It still feels like a really strange API to pass in a structure like this instead of calling `NewMySQL(opts ...Option)`.

containertest/config.go
pdewilde: Ooof, good point. Should I change it back?

containertest/containertest.go
pdewilde: I was matching the existing API, but woudn't this be more extensible?

Anyone can define their own `Service` outside of `containertest` and still use this API.

pdewilde: So if I wanted to make a `Redis` I could do so without having to modify `pkg`

containertest/docker.go
pdewilde: OK, I had made it the concrete type as a hint that it would never be `nil`, is the main consideration just that passing structs by value is too expensive if they grow to be large?

pdewilde: Should MustStart return a pointer too? Or not because its an interface boundary?

containertest/postgres.go
```diff
@@ -0,0 +1,88 @@
...
+	return "postgres"
+}
+
+// ImageTag implements the Service.ImageTag interface.
+func (p *Postgres) ImageTag() string {
+	return p.Version
+}
+
+// TestConn implements the Service.TestConn interface.
+func (p *Postgres) TestConn(progressLogger TestLogger, connInfo ConnInfo) error {
```
sethvargo, line 51: I would expect `ConnInfo` to be a pointer.

sethvargo (approved):

containertest/docker.go
pdewilde: Changed to pointer

containertest/postgres.go
pdewilde: Changed

containertest/config.go
pdewilde: I've changed it to return an error, so could be started within a test rather than just in TestMain. It accepts an interface which is a subset of `testing.TB`.

As a bonus, you could get isolated containers for tests that need them (especially for containers which start quickly)

PR #114, opened 2023-05-02
feat: generalized GitHub app token creation

bradegler:
This allows us to use the same logic for creating installation app tokens from both github-token-minter and github-metrics-aggregator.

### DIFF
--- a/githubapp/githubapp.go
+++ b/githubapp/githubapp.go
@@ -37,20 +37,21 @@ const defaultGitHubAccessTokenURLPattern = "https://api.github.com/app/installat
 
 const cacheKey = "github-app-jwt"
 
-// GitHubAppConfig contains all of the required configuration for operating as a GitHub App.
+// Config contains all of the required configuration for operating as a GitHub App.
 // This includes the three major components, the App ID, the Install ID and the Private Key.
-type GitHubAppConfig struct {
+type Config struct {
 	AppID                 string
 	InstallationID        string
 	PrivateKey            *rsa.PrivateKey
 	accessTokenURLPattern string
 	jwtTokenExpiration    time.Duration
 	jwtCacheDuration      time.Duration
+	client                *http.Client
 }
 
-// GitHubAppConfigOption is a function type that applies a mechanism to set optional
+// ConfigOption is a function type that applies a mechanism to set optional
 // configuration values.
-type GitHubAppConfigOption func(f *GitHubAppConfig)
+type ConfigOption func(f *Config)
 
 // WithAccessTokenURLPattern allows overriding of the GitHub api url that is
 // used when generating installation access tokens. The default is the primary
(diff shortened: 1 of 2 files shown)


githubapp/githubapp.go
```diff
@@ -0,0 +1,229 @@
...
+	InstallationID        string
+	PrivateKey            *rsa.PrivateKey
+	accessTokenURLPattern string
+	jwtTokenExpiration    time.Duration
+	jwtCacheDuration      time.Duration
+}
+
+// GitHubAppConfigOption is a function type that applies a mechanism to set optional
+// configuration values.
+type GitHubAppConfigOption func(f *GitHubAppConfig)
```
sethvargo, line 53: Nit: this name stutters (githubapp.GitHubAppConfig). Maybe name it `Option`?

githubapp/githubapp.go
```diff
@@ -0,0 +1,229 @@
...
+
+	requestReader := bytes.NewReader(requestJSON)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, requestURL, requestReader)
+	if err != nil {
+		return "", fmt.Errorf("error creating http request for GitHub installation information: %w", err)
+	}
+	req.Header.Set("Accept", "application/vnd.github+json")
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", appJWT))
+
+	client := http.Client{Timeout: 10 * time.Second}
```
sethvargo, line 203: Can we make this client during initialization and save it on the struct instead? It might also be good to have a `WithHTTPClient` option.

githubapp/githubapp.go
bradegler: sounds good

githubapp/githubapp.go
```diff
@@ -0,0 +1,229 @@
...
+	if g.jwtCache != nil {
+		// Check for a valid JWT in the cache
+		signedJwt, ok := g.jwtCache.Lookup(cacheKey)
+		if !ok {
+			// Create a JWT for reading instance information from GitHub
+			signedJwt, err := g.generateAppJWT()
+			if err != nil {
+				return nil, fmt.Errorf("error generating the JWT for GitHub app access: %w", err)
+			}
+			g.jwtCache.Set(cacheKey, signedJwt)
```
shan4993, line 144: nit: should we set a variable and do this logic once? i.e. `shouldFetch` or something like that

githubapp/githubapp.go
```diff
@@ -0,0 +1,229 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error generating the JWT for GitHub app access: %w", err)
+	}
+	return signedJwt, nil
+}
+
+// generateAppJWT builds a signed JWT that can be used to
+// communicate with GitHub as an application.
+func (g *GitHubApp) generateAppJWT() ([]byte, error) {
+	// Make the current time 30 seconds in the past to combat clock
```
shan4993, line 159: Should we make the clock skew configurable. Maybe its not worth it, what do you think?

bradegler: I would say if clock skew is more than 30 seconds we've got bigger problems.

githubapp/githubapp.go
bradegler: I renamed GitHubAppConfig to Config, and GitHubAppConfig Option to ConfigOption. I couldn't think of a replacement for GitHubApp but I'm open to it if you have a suggestion.

githubapp/githubapp.go
bradegler: done.

githubapp/githubapp.go
bradegler: ack

sethvargo (approved):

PR #120, opened 2023-05-18
Add a helper for creating a test renderer

sethvargo:


shan4993 (approved):

yolocs:
I wonder if this should live in https://github.com/abcxyz/pkg/tree/main/testutil?

sethvargo:
The renderer is pretty self-contained and might want to be it's own repo/package some day.

Issue #121, opened 2023-05-18
Add flagset for log level in logging package


yolocs:
Per https://github.com/abcxyz/abc/pull/2#discussion_r1198180626

yolocs:
A POC: https://github.com/abcxyz/pkg/pull/126. We want to wait until we officially adopt slog before we introduce this.

sethvargo:
Oh this is done!

PR #124, opened 2023-05-20
feat: add option to initialize logger from flags

yolocs:
Fix: https://github.com/abcxyz/pkg/issues/121

### DIFF
--- a/cli/command_test.go
+++ b/cli/command_test.go
@@ -20,7 +20,6 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/abcxyz/pkg/logging"
 	"github.com/abcxyz/pkg/testutil"
 )
 
@@ -120,7 +119,7 @@ Usage: test COMMAND
 func TestRootCommand_Run(t *testing.T) {
 	t.Parallel()
 
-	ctx := logging.WithLogger(context.Background(), logging.TestLogger(t))
+	ctx := context.Background()
 
 	rootCmd := func() Command {
 		return &RootCommand{
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/abcxyz/pkg/logging"
 	"github.com/abcxyz/pkg/timeutil"
 	"github.com/kr/text"
 )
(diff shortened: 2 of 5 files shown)


cli/command_test.go
```diff
@@ -120,7 +119,7 @@ Usage: test COMMAND
 func TestRootCommand_Run(t *testing.T) {
 	t.Parallel()
 
-	ctx := logging.WithLogger(context.Background(), logging.TestLogger(t))
+	ctx := context.Background()
```
yolocs, line 122: Not really used and it causes circular dependency

sethvargo: This raises a somewhat philosophical question of "should the logger belong to the CLI"... In my head, I was thinking of inverting the problem (making the CLI depend on the logger) so then it would be something like `f.LogLevelFlag(&logger)`.

yolocs: I started with something like that but then I thought it would be nice if the default logger could pick up the flags. Could be over stretching. Will just do the flags then.

yolocs: Tried a different approach. PTAL

logging/logger.go
```diff
@@ -42,15 +42,36 @@ var (
...
 func NewFromEnv(envPrefix string) *zap.SugaredLogger {
 	level := os.Getenv(envPrefix + "LOG_LEVEL")
-	logMode := strings.ToLower(strings.TrimSpace(os.Getenv(envPrefix + "LOG_MODE")))
-	devMode := strings.HasPrefix(logMode, "dev")
+	mode := strings.ToLower(strings.TrimSpace(os.Getenv(envPrefix + "LOG_MODE")))
+	return NewLogger(level + ":" + mode)
+}
+
+// NewLogger creates a new logger from the given config string. The config
+// string is in the format of "level:mode", e.g. "warn:production" or
```
pdewilde, line 58: Why sandwich level and mode together as a string, rather than having two separate arguments?

yolocs: Related to the main change in this PR, I didn't find a more elegant way to initialize a logger with multiple flags. What I did is to coalesce the configs into a single flag (see the flag.go). Because of that, I thought I might as well be consistent to allow creating a logger in the same way. Technically, I can have two arguments here but I dislike the difference between here and the flag.

cli/flags.go
```diff
@@ -685,6 +686,31 @@ func (f *FlagSection) Uint64Var(i *Uint64Var) {
 	})
 }
 
+// LoggerVar is a special flag that binds to a [logging.Logger].
+func (f *FlagSection) LoggerVar(l *logging.Logger) {
```
sethvargo, line 690: I like this design better, but I don't love the devex. I think this is what I really want:

```go
type MyCommand struct {
  logger *logging.Logger
}

func (c *MyCommand) Flags() {
  set := c.NewFlagSet()
  set.AddLoggingFlags(&c.logger)
  return set
}
```

And that would automatically create a section and add multiple flags:

```go
func (s *FlagSet) AddLoggingFlags(l *logging.Logger) {
  f := s.NewSection("LOGGING OPTIONS")

  f.StringVar(&StringVar{
    Name: "log-level",
    EnvVar: "LOG_LEVEL",
    Default: logging.LevelWarn,
    Target: &l.Level,
  })

  f.StringVar(&StringVar{
    Name: "log-format",
    EnvVar: "LOG_FORMAT",
    Default: logging.FormatJSON,
    Target: &l.Format,
  })
}
```

We might end up needing custom getters/setters as you've done here, but I think this is the API we want. WDYT?

yolocs: I don't think I can get this to work without refactoring the logging package. I will close this PR and do the logging refactoring first.

yolocs:
Close to refactor logging first.

PR #125, opened 2023-05-22
Add shell auto-completions support for the CLI

sethvargo:
This adds support for dynamically generating CLI auto-completions (predictions) for command flags and arguments. The default behavior is to predict "something" for all flag values except booleans, and to predict "nothing" for command arguments. This is a superior experience over the current implementation which has no shell completions at all, but developers can further enhance the experience by customizing the CLI flags and commands to predict context-relevant information.

This uses the posener/complete/v2 library, which cleverly self-execs the binary with a special environment that triggers completions. This means completions can be written natively in Go and are automatically versioned with the application. Completions can be as simple as a static list of strings or complex with upstream network calls to predict real-time information.

For more information, see github.com/posener/complete at the master branch.


cli/command_completions_doc_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	f := set.NewSection("Song options")
+
+	f.StringVar(&cli.StringVar{
+		Name:    "song",
+		Aliases: []string{"s"},
+		Example: "Itsy Bitsy Spider",
+		Target:  &c.flagSong,
+		Predict: predict.Set{"Happy Birthday", "Twinkly Twinkle Little Star"},
+		Usage:   "Numeric value by which to increment between each number.",
```
drevell, line 64: This usage string seems inconsistent with the rest of the fields, which are all about songs, not numbers

cli/command_completions_doc_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	f.Int64Var(&cli.Int64Var{
+		Name:    "now",
+		Example: "10404929",
+		Default: time.Now().Unix(),
+		Target:  &c.flagNow,
+		Predict: complete.PredictFunc(func(prefix string) []string {
+			return []string{strconv.FormatInt(time.Now().Unix(), 10)}
+		}),
+		Usage: "Duration to fade audio tracks.",
```
drevell, line 84: Looks like a copy-paste oversight here

cli/command_completions_doc_test.go
```diff
@@ -0,0 +1,125 @@
...
+	// completions is to allow the binary to do it. It will detect the shell and
+	// insert the correct code into the user's shell profile. Instruct users to
+	// run the following command:
+	//
+	//     COMP_INSTALL=1 COMP_YES=1 my-cli
+	//
+	// This will automatically install shell completions. To uninstall
+	// completions, instruct users to run:
+	//
+	//     COMP_UNINSTALL=1 COMP_YES my-cli
```
drevell, line 110: Should COMP_YES be COMP_YES=something?

cli/flags.go
```diff
@@ -233,9 +240,17 @@ type Var[T any] struct {
...
 	Target  *T
 
+	// Parser and Printer are the generic functions for converting string values
+	// to/from the target value. These are populated by the individual flag
+	// helpers.
 	Parser  ParserFunc[T]
 	Printer PrinterFunc[T]
 
+	// Predict is the completion predictor. If no predictor is defined, it
+	// defaults to predicting "something" for all flags except boolean flags.
```
drevell, line 250: Is "something" a technical term of art here? It's not immediately obvious to me what this means.

EDIT: from the below code, I think it means predict.Something, perhaps that could be more clear.

drevell (approved):

cli/flags.go
sethvargo: Done PTAL

cli/command_completions_doc_test.go
sethvargo: Yes, good catch. Fixed.

cli/command_completions_doc_test.go
sethvargo: Thanks, done.

cli/command_completions_doc_test.go
sethvargo: I was having trouble coming up with an example. Fixed the usage.

cli/command.go
```diff
@@ -27,8 +27,16 @@ import (
 	"strings"
 
 	"github.com/mattn/go-isatty"
+	"github.com/posener/complete/v2"
+	"github.com/posener/complete/v2/predict"
 )
 
+// isCompletionRequest returns true if the invocation is a completion request,
+// or false otherwise.
```
yolocs, line 35: nit: could you expand the comment the say these are env vars recognized by [github.com/posener](https://github.com/posener/complete)? It was not obvious to me initially.

yolocs (approved):

cli/command_completions_doc_test.go
```diff
@@ -0,0 +1,125 @@
...
+	// see even more ways, look at the examples at
+	// github.com/posener/complete/tree/master.
+	//
+	// Since completions are a shell function, users must add something to their
+	// shell configuration (e.g. .bashrc, .zshrc). The easiest method to install
+	// completions is to allow the binary to do it. It will detect the shell and
+	// insert the correct code into the user's shell profile. Instruct users to
+	// run the following command:
+	//
+	//     COMP_INSTALL=1 COMP_YES=1 my-cli
```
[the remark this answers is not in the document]
sethvargo: https://github.com/posener/complete/blob/3f9152130d1c1e72ef5b0091380bfbeb7fafecf5/complete.go#L61-L65

cli/command.go
sethvargo: Done.

PR #126, opened 2023-05-22
POC logger factory

yolocs:
Challenges

To have something like described in https://github.com/abcxyz/pkg/pull/124#discussion_r1201122470, we need setters on the **logger** to update level or encoding. However, `zap` doesn't provide any setters to change log level or encoding on an existing logger (same as `slog`). To have a logger with desired level or encoding, we have create a new logger.

First I tried to create a wrapper around the native logger, like

```go
type Logger struct {
  *zap.SugaredLogger
  Options *Options
}

type Options struct {
  Level, Mode string
}
```

This worked poorly because we need to change logger type _everywhere_ it's used. Considering we will try to move to `slog` once it's out of `exp`, I don't think it worths the effort to do so.

So instead, I tried the way in this PR. Instead of initializing a logger directly from flags, we initialize a logger factory and use the factory to create loggers. I think this would work for `slog` as well.

### DIFF
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -300,8 +300,8 @@ func Flag[T any](f *FlagSection, i *Var[T]) {
 		}
 	}
 
-	// Set a default value.
-	*i.Target = initial
+	// Use the setter to set the default value.
+	setter(i.Target, initial)
 
 	// Compute a sane default if one was not given.
 	example := i.Example
@@ -731,6 +731,48 @@ func (f *FlagSection) Uint64Var(i *Uint64Var) {
 		Target:  i.Target,
 		Parser:  parser,
 		Printer: printer,
+	})
+}
+
+// LoggerConfigurer is the interface to configure a logger.
+type LoggerConfigurer interface {
+	SetLevel(level string)
+	SetMode(mode string)
+}
+
+// LoggerVar binds common logging flags to a [LoggerConfigurer].
+func (f *FlagSection) LoggerVar(i LoggerConfigurer) {
+	parser := func(s string) (string, error) { return s, nil }
+	printer := func(v string) string { return v }
+	var dummy string
+
+	Flag(f, &Var[string]{
+		Name:    "log-level",
+		Usage:   "To specify the log level with one of debug|info|warn|error|fatal",
+		Example: "info",
+		Default: "warn",
+		EnvVar:  "LOG_LEVEL",
+		Predict: predict.Set{"debug", "info", "warn", "error", "fatal"},
+		Target:  &dummy,
+		Parser:  parser,
+		Printer: printer,
+		Setter: func(cur *string, val string) {
+			i.SetLevel(val)
+		},
+	})
+	Flag(f, &Var[string]{
+		Name:    "log-mode",
+		Usage:   "To specify the log mode with one of dev|production",
+		Example: "production",
+		Default: "production",
+		EnvVar:  "LOG_MODE",
+		Predict: predict.Set{"dev", "production"},
+		Target:  &dummy,
+		Parser:  parser,
+		Printer: printer,
+		Setter: func(cur *string, val string) {
+			i.SetMode(val)
+		},
 	})
 }
 
--- a/cli/flags_test.go
+++ b/cli/flags_test.go
@@ -20,6 +20,9 @@ import (
 	"reflect"
 	"strings"
 	"testing"
+
+	"github.com/abcxyz/pkg/logging"
+	"github.com/google/go-cmp/cmp"
 )
 
 func TestNewFlagSet(t *testing.T) {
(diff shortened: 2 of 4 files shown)


yolocs:
cc @sethvargo POC for discussion. My thinking process is in the description

cli/flags.go
```diff
@@ -300,8 +300,8 @@ func Flag[T any](f *FlagSection, i *Var[T]) {
 		}
 	}
 
-	// Set a default value.
-	*i.Target = initial
+	// Use the setter to set the default value.
+	setter(i.Target, initial)
```
yolocs, line 304: This was a bug. If the setter is used to set value, the original logic won't trigger it to set the default value.

cli/flags.go
```diff
@@ -734,6 +734,48 @@ func (f *FlagSection) Uint64Var(i *Uint64Var) {
 	})
 }
 
+// LoggerConfigurer is the interface to configure a logger.
+type LoggerConfigurer interface {
```
yolocs, line 738: I think this is better since it decouples the logger from the cli package.

cli/flags_test.go
```diff
@@ -91,6 +94,51 @@ func TestFlagSet_Help(t *testing.T) {
 	}
 }
 
+func TestLoggerVar(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name        string
+		args        []string
+		wantFactory *logging.Factory
```
yolocs, line 103: To truly decouple from the logging package, I could use a local fake instead.

logging/logger.go
```diff
@@ -42,15 +42,43 @@ var (
 	defaultLoggerOnce sync.Once
 )
 
+// Factory is a factory to create loggers.
+type Factory struct {
```
yolocs, line 46: To use the existing logging package with the CLI flags, one would need to use this factory instead.

```go
lf := &logging.Factory{}
fs.LoggerVar(lf)
logger := lf.New()
```

logging/logger.go
```diff
@@ -42,15 +42,43 @@ var (
...
+	Level, Mode string
+}
+
+// SetLevel sets the level for the logger factory.
+func (f *Factory) SetLevel(level string) {
```
pdewilde, line 51: I know the fluent pattern isn't loved in golang, but this seems like the ideal place for it. Why isn't it liked?

yolocs: You mean like `WithLevel(level string) *Factory`?

pdewilde: yeah

yolocs: I was actually trying to do that. However, it doesn't work very well with Go interface. E.g.

```go
type LoggerConfigurer interface {
  WithLevel(level string) LoggerConfigurer
}

type Factory struct {}

// This doesn't implement the interface above because of the return value type.
// But it's also kinda bad to return LoggerConfigurer instead.
func (f *Factory) WithLevel(level string) *Factory {}
```

cli/flags.go
sethvargo: I agree on the interface approach since it decouples. It feels a little weird to have it embedded in the flags package. Maybe we want to revisit changing the test to remove the circular dependency if we're going to go the interface route? Idk - it feels like we need to pick the best of the worst.

cli/flags.go
```diff
@@ -734,6 +734,48 @@ func (f *FlagSection) Uint64Var(i *Uint64Var) {
...
+	printer := func(v string) string { return v }
+	var dummy string
+
+	Flag(f, &Var[string]{
+		Name:    "log-level",
+		Usage:   "To specify the log level with one of debug|info|warn|error|fatal",
+		Example: "info",
+		Default: "warn",
+		EnvVar:  "LOG_LEVEL",
+		Predict: predict.Set{"debug", "info", "warn", "error", "fatal"},
```
sethvargo, line 755: Not for this PR, but as we think about the migration to slog, it would be nice if these value came from constants in the logging package.

cli/flags.go
```diff
@@ -734,6 +734,48 @@ func (f *FlagSection) Uint64Var(i *Uint64Var) {
...
+
+// LoggerVar binds common logging flags to a [LoggerConfigurer].
+func (f *FlagSection) LoggerVar(i LoggerConfigurer) {
+	parser := func(s string) (string, error) { return s, nil }
+	printer := func(v string) string { return v }
+	var dummy string
+
+	Flag(f, &Var[string]{
+		Name:    "log-level",
+		Usage:   "To specify the log level with one of debug|info|warn|error|fatal",
```
sethvargo, line 751: Nit: it might be better to do something like:

```go
levels := []string{"debug", "info", "warn", "error", "fatal"}

//
Usage: "To specify the log level with one of " + strings.Join(levels, "|")
//
Predict: predict.Set(levels)
```

logging/logger.go
sethvargo: I'm not sure I follow why we need this?

cli/flags.go
yolocs: Sent a separate PR

logging/logger.go
yolocs: `zap.SugaredLogger` doesn't provide a way to set level or mode dynamically (see my thinking in the PR description). A factory is the next best thing I could thought of

sethvargo: Ah I see. What about when we move to slog? We might want to wait the ~month until 1.21 is released and just do this once with the slog package instead of trying to do something more general.

yolocs: sg! Let me comment in the slog PR so don't forget.

PR #134, opened 2023-06-06
Enhancement: Add ExposedPorts to container setup RunOptions

kevya-google:
Adding this change replicates the behavior for running with `docker run -p` port mapping, prior to adding this, containertest relies on the Dockerfile declaring exposed ports.


drevell:
> Would it be possible to test this?

I can't think of a good way.

PR #135, opened 2023-06-09
Introduce multicloser

yolocs:
See package comment about usage. 
TODO: add unit test

### DIFF
--- a/multicloser/multicloser.go
+++ b/multicloser/multicloser.go
@@ -39,7 +39,7 @@
 //
 //		client2 , err := newClient()
 //		if err != nil {
-//			return closer, err
+//			return closer, err // client1 will still be closed in this case.
 //		}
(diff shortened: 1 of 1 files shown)


sethvargo (commented):
I went down this path, but I found that the return types for closers are usually `void` or `error`. I was trying to find a way to accept varying type signatures with generics, but failed. The design I had come up with was:

```go
var mc multicloser.Closer

kmsClient, err := thing.New()
if err != nil {
  return mc, err
}
mc.Do(kmsClient.Close)

mc.DoVoid(otherthing.Close)
```

and `DoVoid` just wrapped the function in a closure that returned `nil`.

I do think we want to return an error, so we match the io.Closer interface: https://pkg.go.dev/io#Closer

yolocs:
Good idea. Will try that.

PR #136, opened 2023-06-10
Add multicloser implementation

sethvargo:
Alternative approach to https://github.com/abcxyz/pkg/pull/135

### DIFF
--- a/multicloser/multicloser.go
+++ b/multicloser/multicloser.go
@@ -1,0 +1,87 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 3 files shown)


multicloser/multicloser.go
```diff
@@ -0,0 +1,87 @@
...
+//
+// It is not safe to use concurrently without locking.
+type Closer struct {
+	fns []func() error
+}
+
+// Append adds the given closer functions. It handles void and error signatures.
+// Other signatures should use an anonymous function to match an expected
+// signature.
+func Append[T Func](c *Closer, fns ...T) *Closer {
```
sethvargo, line 42: This works, and it's typesafe, but you can't _blend types_ in the same call.

```go
// Good
c = multicloser.Append(c, func() {}, func() {})

// Good
c = multicloser.Append(c, func() error {}, func() error {})

// BAD!
c = multicloser.Append(c, func() {}, func() error {})
```

multicloser/multicloser.go
```diff
@@ -0,0 +1,87 @@
...
+package multicloser
+
+import (
+	"errors"
+)
+
+// Func is the type signature for a closing function. It accepts a function that
+// returns an error or a void function.
+type Func interface {
+	func() error | func()
```
sethvargo, line 28: You **cannot** make these concrete types or the type system freaks out. This is the situation I was running into the first time I tried.

yolocs (approved):
You beat me! Ship it!

multicloser/multicloser.go
drevell: One possibility: add function WrapVoid function that takes a `func() void` and returns a no-op `func() error`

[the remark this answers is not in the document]
sethvargo: https://github.com/abcxyz/pkg/pull/137

PR #138, opened 2023-06-13
logging: switch to using log/slog

sethvargo:


sethvargo:
Looks like golangci-lint still isn't quite happy...

logging/logger.go
```diff
@@ -12,179 +12,179 @@
...
 	if err != nil {
-		logger = zap.NewNop()
+		if debugEnvVarValue != "" {
+			panic(fmt.Sprintf("invalid value for %s (%s): %s",
+				debugEnvVarKey, debugEnvVarValue, err))
+		}
 	}
 
-	return logger.Sugar()
+	return New(os.Stderr, level, format, debug)
```
yolocs, line 128: Why `os.Stderr` here?

logging/logger.go
```diff
@@ -12,179 +12,179 @@
...
-	// include upon calling DefaultLogger.
-	defaultLogger     *zap.SugaredLogger
-	defaultLoggerOnce sync.Once
-)
+// defaultLogger returns a function that returns the default logger. which
+// writes JSON output to stdout at the "Info" level.
+//
+// It is initialized once when called the first time.
+var defaultLoggerOnce = sync.OnceValue[*slog.Logger](func() *slog.Logger {
+	v, _ := New(os.Stdout, LevelInfo, FormatJSON, false)
```
yolocs, line 48: Should we use `NewFromEnv` for the default? One can simply use env vars to set up the default logger.

logging/logger.go
sethvargo: Bug

logging/logger.go
sethvargo: I think it was a mistake that the default logger parsed the environment. `NewFromEnv` can also panic when given invalid input, which seems like a bad idea for the default

yolocs:
> @yolocs okay so for rollout plan here, since this contains breaking changes, I think we should:
> 
> 1. Cut a new release of pkg
> 2. Merge this
> 3. Cut a new release of pkg and note the breaking changes
> 4. Start opening PRs to update deps and change the syntax where needed
> 
> Thoughts?

sgtm, along (3) give team a headsup

yolocs (approved):
LGTMing to move forward

PR #142, opened 2023-06-23
Add override-workflow-variables action

gjonathanhong:
This is a reusable GitHub action for loading env variables from an env file of the working-directory, if it exists, into the `GITHUB_ENV` environment file. 

To require the use of an env file, set `require_env = 'true'`  and optionally specify `enforced_directories` (in CSV string) to enforce upon specific directories and their subdirectories. This is useful when you need separated env variables between directories.

### DIFF
--- a/.github/actions/override-workflow-variables/action.yml
+++ b/.github/actions/override-workflow-variables/action.yml
@@ -26,7 +26,7 @@ inputs:
     required: false
     default: '.env'
   require_env:
-    description: 'fail if the env file isn't found; unless enforced_directories is set, every working_directory must have an .env file'
+    description: 'fail if the env file isn't found; unless enforced_directories is set, every working_directory must have an .env file.'
     required: false
     default: 'false'
   enforced_directories:
--- a/.github/actions/override-workflow-variables/action.yml
+++ b/.github/actions/override-workflow-variables/action.yml
@@ -26,7 +26,7 @@ inputs:
     required: false
     default: '.env'
   require_env:
-    description: 'fail if the env file isn't found; unless enforced_directories is set, every working_directory must have an .env file.'
+    description: 'fail if the env file is not found; unless enforced_directories is set, every working_directory must have an .env file.'
     required: false
     default: 'false'
   enforced_directories:
--- a/.github/actions/override-workflow-variables/action.yml
+++ b/.github/actions/override-workflow-variables/action.yml
@@ -30,7 +30,7 @@ inputs:
     required: false
     default: 'false'
   enforced_directories:
-    description: 'only used when require_env=true; rather than looking for an env file in every immediate subdirectory, just require it in this set of directories.'
+    description: 'only used when require_env=true; rather than looking for an env file in all working-directories, just require it in working-directories belonging to this set of parent directories.'
     required: false
     default: ''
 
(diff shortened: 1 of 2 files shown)


.github/actions/override-workflow-variables/action.yml
```diff
@@ -0,0 +1,68 @@
...
+  working_directory:
+    description: 'The working directory to look for the env file.'
+    required: false
+    default: '.'
+  env_file:
+    description: 'The env file containing variables to override with.'
+    required: false
+    default: '.env'
+  require_env:
+    description: 'Require the use of an env file for overriding variables.'
```
drevell, line 29: I found this a bit hard to understand, maybe consider something like `fail if the env file isn't found; unless enforced_directories is set, each subdirectory (non-recursively) of the working_directory must have an .env file`

.github/actions/override-workflow-variables/action.yml
```diff
@@ -0,0 +1,68 @@
...
+    required: false
+    default: 'false'
+  enforced_directories:
+    description: 'A CSV string of directories to require the use of an env file for overriding variables.'
```
drevell, line 33: I also had a hard time understanding this, maybe considering something like `only used when require_env=true; rather than looking for an env file in every immediate subdirectory, just require it in this set of directories`.

.github/actions/override-workflow-variables/action.yml
```diff
@@ -0,0 +1,68 @@
...
+        REQUIRE_ENV: '${{ inputs.require_env }}'
+        ENFORCED_DIRECTORIES: '${{ inputs.enforced_directories }}'
+      run: |-
+        if [ -f "$ENV_FILE" ]; then
+          # Filter out comment lines, empty lines, and lines with only whitespace.
+          grep -v '^[[:space:]]*#\|^$\|^\s*$' $ENV_FILE >> ${GITHUB_ENV}
+        elif [ "$REQUIRE_ENV" == true ]; then
+
+          if [ "$ENFORCED_DIRECTORIES" == "" ]; then
+            # Set to top-level dir of respository to enforce on all subdirectories.
```
drevell, line 54: Maybe make it clear that "all subdirectories" isn't recursive, it's just immediate children

.github/actions/override-workflow-variables/action.yml
```diff
@@ -0,0 +1,68 @@
...
+inputs:
+  working_directory:
+    description: 'The working directory to look for the env file.'
+    required: false
+    default: '.'
+  env_file:
+    description: 'The env file containing variables to override with.'
+    required: false
+    default: '.env'
+  require_env:
```
drevell, line 28: Can you help me understand what this is for / why it's needed? I understand the point of loading an .env file, but I don't understand the reason for the the other stuff to enforce the presence of files that are never loaded.

drevell (changes requested):

.github/actions/override-workflow-variables/action.yml
gjonathanhong: I changed the wording from subdirectories to working-directories.

e.g. If the `enforced_directories = 'foo/'`, then a working directory of `foo/bar/baz` and `foo/bar` will require `foo/bar/baz/.env` and `foo/bar/.env` respectively.

Does this make it clearer?

.github/actions/override-workflow-variables/action.yml
verbanicm: I believe the idea is that you may want to always load an env file for a specific workflow and this would prevent misconfiguration. The specific use case was guardian, if we have environment folders, we want to make sure each environment has its own file.

gjonathanhong: This is useful for situations when workflow variables with the same name are reused across different folders. If two folders within a repository belong to two separate product teams, we may need to enforce `require_env=true` on their parent folders to ensure they are using separate service accounts to perform actions for their product only. 

in short, we can share the same workflow but keep separation between product folders/teams.

drevell (approved):

verbanicm (approved):

PR #143, opened 2023-06-25
fix: allow cli prompt to be canceled

verbanicm:
When adding prompt feature to https://github.com/abcxyz/abc, I noticed that prompt cannot be canceled with SIGTERM or SIGINT (even it parent context was created that way). This adds that capability.

Chose to add context to prompt func to allow calling function to determine how the context should be canceled in addition to adding signal.NotifyContext to guarantee Prompt is cancelable.

### DIFF
--- a/cli/command.go
+++ b/cli/command.go
@@ -279,6 +279,9 @@ func (c *BaseCommand) Prompt(ctx context.Context, msg string, args ...any) (stri
 	errCh := make(chan error, 1)
 
 	// Prompts for user input and returns the value or an error if there is one.
+	// If the context is canceled, this function leaves the c.Stdin in a bad state.
+	// This is currently acceptable as the cli process should ideallly be terminated
+	// after the context is canceled.
 	go func() {
 		defer close(resultCh)
 		defer close(errCh)
--- a/cli/command.go
+++ b/cli/command.go
@@ -63,10 +63,7 @@ type Command interface {
 	// Run executes the command.
 	Run(ctx context.Context, args []string) error
 
-	// Prompt provides a mechanism for asking for user input. It reads from
-	// [Stdin]. If there's an input stream (e.g. a pipe), it will read the pipe.
-	// If the terminal is a TTY, it will prompt. Otherwise it will fail if there's
-	// no pipe and the terminal is not a tty.
+	// Prompt provides a mechanism for asking for user input.
 	Prompt(ctx context.Context, msg string, args ...any) (string, error)
 
 	// Stdout returns the stdout stream. SetStdout sets the stdout stream.
@@ -272,51 +269,33 @@ func (c *BaseCommand) Hidden() bool {
 	return false
 }
 
-// Prompt prompts the user for a value. If stdin is a tty, it prompts. Otherwise
-// it reads from the reader.
+// Prompt prompts the user for a value. It reads from [Stdin], up to 64k bytes.
+// If there's an input stream (e.g. a pipe), it will read the pipe.
+// If the terminal is a TTY, it will prompt. Otherwise it will fail if there's
+// no pipe and the terminal is not a tty. If the context is canceled, this function
+// leaves the c.Stdin in a bad state.
 func (c *BaseCommand) Prompt(ctx context.Context, msg string, args ...any) (string, error) {
-	resultCh := make(chan string, 1)
-	errCh := make(chan error, 1)
+	if c.Stdin() == os.Stdin && isatty.IsTerminal(os.Stdin.Fd()) {
+		fmt.Fprint(c.Stdout(), msg, args)
+	}
 
-	// Prompts for user input and returns the value or an error if there is one.
-	// If the context is canceled, this function leaves the c.Stdin in a bad state.
-	// This is currently acceptable as the cli process should ideallly be terminated
-	// after the context is canceled.
+	scanner := bufio.NewScanner(io.LimitReader(c.Stdin(), 64*1_000))
+	finished := make(chan struct{})
 	go func() {
-		defer close(resultCh)
-		defer close(errCh)
-
-		scanner := bufio.NewScanner(io.LimitReader(c.Stdin(), 64*1_000))
-
-		if c.Stdin() == os.Stdin && isatty.IsTerminal(os.Stdin.Fd()) {
-			fmt.Fprint(c.Stdout(), msg, args)
-		}
-
+		defer close(finished)
 		scanner.Scan()
-
-		if err := scanner.Err(); err != nil {
-			select {
-			case errCh <- fmt.Errorf("failed to read stdin: %w", err):
-			default:
-			}
-		}
-
-		select {
-		case resultCh <- scanner.Text():
-		default:
-		}
 	}()
 
-	// If input is provided and the provided context are cancelled simultaneously,
-	// the behavior is undefined. See https://www.sethvargo.com/what-id-like-to-see-in-go-2/#deterministic-select.
 	select {
 	case <-ctx.Done():
 		return "", fmt.Errorf("failed to prompt: %w", ctx.Err())
-	case err := <-errCh:
-		return "", fmt.Errorf("failed to prompt: %w", err)
-	case result := <-resultCh:
-		return result, nil
+	case <-finished:
 	}
+
+	if err := scanner.Err(); err != nil {
+		return "", fmt.Errorf("failed to read stdin: %w", err)
+	}
+	return scanner.Text(), nil
 }
 
 // Outf is a shortcut to write to [BaseCommand.Stdout].
--- a/cli/command.go
+++ b/cli/command.go
@@ -280,7 +280,8 @@ func (c *BaseCommand) Hidden() bool {
 // it reads from the reader.
 func (c *BaseCommand) Prompt(ctx context.Context, msg string, defaultValue *string) (string, error) {
 	// guarantee Prompt is cancelable with SIGTERM and SIGINT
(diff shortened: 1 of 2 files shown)


sethvargo (changes requested):
I'm not sure this is needed? Does it actually fail to cancel on ctrl+c?

cli/command.go
```diff
@@ -274,19 +278,42 @@ func (c *BaseCommand) Hidden() bool {
...
 
-	if err := scanner.Err(); err != nil {
-		return "", fmt.Errorf("failed to read stdin: %w", err)
+	select {
+	case <-ctx.Done():
+		return "", fmt.Errorf("failed to prompt: %w", ctx.Err())
+	case err := <-errorChan:
+		return "", fmt.Errorf("failed to prompt: %w", err)
+	case result := <-resultChan:
+		return result, nil
```
sethvargo, line 315: Like we discussed, this doesn't do what you think it does.

cli/command.go
```diff
@@ -274,19 +278,42 @@ func (c *BaseCommand) Hidden() bool {
...
+			fmt.Fprint(c.Stdout(), msg)
+		}
+
+		scanner.Scan()
+
+		if err := scanner.Err(); err != nil {
+			errorChan <- fmt.Errorf("failed to read stdin: %w", err)
+		}
+
+		if defaultValue != nil && strings.TrimSpace(scanner.Text()) == "" {
```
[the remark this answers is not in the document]
verbanicm: Because `scanner.Text()` will return "" for no input and it seemed silly to set "" = "" if no default value is given. I suppose it doesnt matter really tho.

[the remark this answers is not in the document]
verbanicm: Fair, shouldnt force that.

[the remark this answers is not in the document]
verbanicm: Should we just remove defaultValue and let the user do it on their end?

cli/command.go
verbanicm: Ill leave a comment.

cli/command.go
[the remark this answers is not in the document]
verbanicm: Removing default value

cli/command.go
```diff
@@ -274,19 +274,46 @@ func (c *BaseCommand) Hidden() bool {
 
 // Prompt prompts the user for a value. If stdin is a tty, it prompts. Otherwise
 // it reads from the reader.
-func (c *BaseCommand) Prompt(msg string) (string, error) {
-	scanner := bufio.NewScanner(io.LimitReader(c.Stdin(), 64*1_000))
+func (c *BaseCommand) Prompt(ctx context.Context, msg string, args ...any) (string, error) {
```
drevell, line 277: If canceled, this function leaks a goroutine that is reading from stdin. That could be a problem if there are any further reads from stdin. Later reads will behave weirdly since the read they're trying to do might get stolen by this orphaned goroutine.

Suggestions:
 - document this behavior and leave it as-is. Say something like "if the context is canceled, this function leaves stdin in a bad state."
 - Interrupt the background goroutine reading from stdin by closing stdin (or some other way of using Close()) to interrupt the background goroutine. Not sure if this works.
 - Somehow implement a cancelable io.Reader in pkg

cli/command.go
```diff
@@ -274,19 +274,46 @@ func (c *BaseCommand) Hidden() bool {
...
-	if err := scanner.Err(); err != nil {
-		return "", fmt.Errorf("failed to read stdin: %w", err)
+		if err := scanner.Err(); err != nil {
+			select {
+			case errCh <- fmt.Errorf("failed to read stdin: %w", err):
+			default:
+			}
+		}
+
+		select {
```
drevell, line 301: Can you help me understand why we need `select` here? Can it be just `resultCh <- scanner.Text()`?

drevell (changes requested):

cli/command.go
verbanicm: See comment from Seth: https://github.com/abcxyz/pkg/pull/143#discussion_r1242225484

cli/command.go
verbanicm: Added defer close on os.Stdin

verbanicm: Changed to comment on the issue below

cli/command.go
```diff
@@ -274,19 +274,49 @@ func (c *BaseCommand) Hidden() bool {
...
 // Prompt prompts the user for a value. If stdin is a tty, it prompts. Otherwise
 // it reads from the reader.
-func (c *BaseCommand) Prompt(msg string) (string, error) {
-	scanner := bufio.NewScanner(io.LimitReader(c.Stdin(), 64*1_000))
+func (c *BaseCommand) Prompt(ctx context.Context, msg string, args ...any) (string, error) {
+	resultCh := make(chan string, 1)
+	errCh := make(chan error, 1)
+
+	// Prompts for user input and returns the value or an error if there is one.
+	// If the context is canceled, this function leaves the c.Stdin in a bad state.
```
drevell, line 282: I think this sentence (`If the context is canceled, this function leaves the c.Stdin in a bad state.`) should go in the godoc, in the comment preceding this function declaration.

cli/command.go
drevell: But the channel is buffered, so it can't block. The channel's capacity is 1, and there's only ever 1 send, so no blocking can happen. The select can be removed.

But if you take my suggestion above for a simpler way to write this function, this whole thing goes away.

cli/command.go
```diff
@@ -274,19 +278,42 @@ func (c *BaseCommand) Hidden() bool {
...
+	defer done()
 
-	if c.Stdin() == os.Stdin && isatty.IsTerminal(os.Stdin.Fd()) {
-		fmt.Fprint(c.Stdout(), msg)
-	}
+	resultChan := make(chan string)
+	errorChan := make(chan error)
+
+	go func() {
+		scanner := bufio.NewScanner(io.LimitReader(c.Stdin(), 64*1_000))
```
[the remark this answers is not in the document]
sethvargo: Sorry, I'm still not seeing the 64k read limit documented in the Godoc for `Prompt`.

cli/command.go
```diff
@@ -274,19 +274,49 @@ func (c *BaseCommand) Hidden() bool {
...
+			fmt.Fprint(c.Stdout(), msg, args)
+		}
 
-	if c.Stdin() == os.Stdin && isatty.IsTerminal(os.Stdin.Fd()) {
-		fmt.Fprint(c.Stdout(), msg)
-	}
+		scanner.Scan()
 
-	scanner.Scan()
+		if err := scanner.Err(); err != nil {
```
drevell, line 297: I think this whole function can be simplified:

```
func (...) Prompt(...) string {
  if c.Stdin() == os.Stdin && isatty.IsTerminal(os.Stdin.Fd()) {
    fmt.Fprint(c.Stdout(), msg, args)
  }
  scanner := bufio.NewScanner(io.LimitReader(c.Stdin(), 64*1_000)
  finished := make(chan struct{})
  go func() {
    scanner.Scan()
    close(finished)
  }()
  
  select {
    case <-ctx.Done():
      return "", fmt.Errorf("failed to prompt: %w", ctx.Err())
    case <-finished:
  }
  
  if err := scanner.Err(); err != nil {
    return fmt.Errorf("failed to read stdin: %w", err)
  }
  return scanner.Text(), nil
}
```

cli/command_test.go
```diff
@@ -269,6 +269,125 @@ func TestRootCommand_Run(t *testing.T) {
 	}
 }
 
+func TestRootCommand_Prompt_Values(t *testing.T) {
```
sethvargo, line 272:
```suggestion
func TestBaseCommand_Prompt_Values(t *testing.T) {
```

cli/command_test.go
```diff
@@ -269,6 +269,125 @@ func TestRootCommand_Run(t *testing.T) {
...
+				}
+				if got, want := strings.TrimSpace(stderr.String()), strings.TrimSpace(tc.expStderr); !strings.Contains(got, want) {
+					t.Errorf("expected\n\n%s\n\nto contain\n\n%s\n\n", got, want)
+				}
+			}
+		})
+	}
+}
+
+func TestRootCommand_Prompt_Cancels(t *testing.T) {
```
sethvargo, line 329:
```suggestion
func TestBaseCommand_Prompt_Cancels(t *testing.T) {
```

drevell (changes requested):

cli/command.go
sethvargo: I'd probably do:

```
  go func() {
    defer close(finished)
    scanner.Scan()
  }()
```

but yea I think you're right

verbanicm: Thanks for the input!

cli/command.go
verbanicm: Moved.

cli/command.go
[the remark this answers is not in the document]
verbanicm: Added

cli/command.go
```diff
@@ -63,11 +63,8 @@ type Command interface {
 	// Run executes the command.
 	Run(ctx context.Context, args []string) error
 
-	// Prompt provides a mechanism for asking for user input. It reads from
-	// [Stdin]. If there's an input stream (e.g. a pipe), it will read the pipe.
-	// If the terminal is a TTY, it will prompt. Otherwise it will fail if there's
-	// no pipe and the terminal is not a tty.
-	Prompt(msg string) (string, error)
+	// Prompt provides a mechanism for asking for user input.
```
verbanicm, line 66: Shortened and moved this verbiage to the implementation method below.

sethvargo (approved):

PR #146, opened 2023-06-29
feat: programmatic fallback log level and mode

drevell:
This adds the function `logging.NewFromEnvWithDefaults()`. This allows the program to provide a log level and a mode (dev/prod) that will be used in the case that the _LOG_LEVEL and/or _LOG_MODE env vars aren't set.

The initial use case is for the abc CLI, where we want to use the "dev" log mode unless the user specifically requested otherwise by setting an env var.

### DIFF
--- a/logging/logger.go
+++ b/logging/logger.go
@@ -42,12 +42,51 @@ var (
 	defaultLoggerOnce sync.Once
 )
 
+// FallbackConfig specifies a logging configuration to use if environment
+// variables aren't set.
+type FallbackConfig struct {
+	// The log mode to use if the environment variable isn't set. Valid values
+	// are "dev" and "production".
+	ModeIfNoEnv string
+
+	// The log level to use if the environment variable isn't set. Valid values
+	// are anything accepted by zap (debug, info, warn, error, dpanic, panic,
+	// fatal).
+	LevelIfNoEnv string
+}
+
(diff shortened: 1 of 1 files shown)


sethvargo (commented):
What's the value of this over the following in a program:

```go
if os.Getenv("LOG_LEVEL") == "" {
  os.Setenv("LOG_LEVEL", "dev")
}
```

When we switch to slog, some of the notion of "mode" will go away. Mode is a very Zap-centric thing, so I'm hesitant to couple to it.

drevell:
> What's the value of this over the following in a program:
> 
> ```go
> if os.Getenv("LOG_LEVEL") == "" {
>   os.Setenv("LOG_LEVEL", "dev")
> }
> ```
> 
> When we switch to slog, some of the notion of "mode" will go away. Mode is a very Zap-centric thing, so I'm hesitant to couple to it.

I'd be fine with that, it just seemed a bit hacky to pass parameters via environment vars. The new API in this PR was the consensus from an old thread in the team chat (timestamp May 31, 11:30 AM).

I'd be fine with just using the env var to avoid adding a new API to something that's going away. I'll close this PR until/unless anybody expresses an opinion otherwise.

sethvargo:
Thanks - that sounds good. Once 1.21 comes out (~1 month), we can think about adding this to the slogger wrapper.

Issue #148, opened 2023-07-10
workerpool: return multierror in Done


sethvargo:
### TL;DR

Create an ordered multierror to return after `Done()`:
 https://github.com/abcxyz/pkg/blob/f75fec6a9ad6e1b5d6b5a8fbb494f28644ad18af/worker/worker.go#L158-L163

PR #151, opened 2023-07-10
Return a multi-error in Done

sethvargo:
This changes the API contract for Done(). Previously it would only return an error if the context was cancelled or the worker pool was stopped. Now it also returns an error that includes a wrapped error that contains all of the errors of the tasks.

Fixes https://github.com/abcxyz/pkg/issues/148


workerpool/workerpool_test.go
```diff
@@ -116,7 +116,10 @@ func TestWorker_Done(t *testing.T) {
 			got = append(got, fmt.Sprintf("%s", result.Error))
 		}
 		if diff := cmp.Diff(want, got); diff != "" {
-			t.Errorf("justs: diff (-want, +got):\n%s", diff)
+			t.Error(diff)
+		}
+		if diff := cmp.Diff(want, got); diff != "" {
+			t.Error(diff)
```
verbanicm, line 122: Am i seeing double?

verbanicm (approved):

PR #156, opened 2023-07-13
workerpool: add stop-on-error functionality

sethvargo:
This is a breaking API change, mostly because the "option pattern" with Go doesn't work well with generics. The Option has to carry the type signature (which is fine), but the call site requires specifying the type signature again, and it felt really bad. I'd like:

```go
pool := workerpool.New[string](3, workerpool.WithStopOnError(true))
```

but the Option also needs a type inference:

```go
pool := workerpool.New[string](3, workerpool.WithStopOnError[string](true))
```

And that would need to be repeated for each option...

### DIFF
--- a/workerpool/workerpool.go
+++ b/workerpool/workerpool.go
@@ -45,6 +45,8 @@ type Pool[T any] struct {
 	results     []*result[T]
 	resultsLock sync.Mutex
 
+	stopOnError bool
+
 	stopped uint32
 }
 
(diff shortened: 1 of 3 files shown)


drevell (approved):

workerpool/workerpool.go
```diff
@@ -100,37 +119,35 @@ func New[T any](concurrency int64) *Pool[T] {
 //
 // Never call Do from within a Do function because it will deadlock.
 func (p *Pool[T]) Do(ctx context.Context, fn WorkFunc[T]) error {
-	// Do not enqueue new work if the worker pool is stopped.
+	i := atomic.AddInt64(&p.i, 1)
+
 	if p.isStopped() {
+		p.appendResult(i, *new(T), ErrStopped)
 		return ErrStopped
```
dcreey, line 126: Should we consider removing the error response from this func since we will get all the errors when Done is called? That will be a bit nicer on the consumer side.

sethvargo: You might want to stop queuing new work on the client based on the fact that the worker has stopped.

dcreey: Ack

PR #165, opened 2023-07-21
sets: introduce sets package for common set-like operations

sethvargo:
This introduces some library functions that are not present in `exp/slices` and `exp/maps` today, notably set intersection, union, and subtraction operations.

The package exposes both `slice` and `map` variants, but the `map` functions are named `*MapKeys` to emphasize that set operations are performed on the keys as opposed to the values.

### DIFF
--- a/sets/maps.go
+++ b/sets/maps.go
@@ -123,6 +123,11 @@ func SubtractMapKeys[K comparable, V any](maps ...map[K]V) map[K]V {
 	}
 
 	for _, m := range maps[1:] {
+		// Short-circuit: if the map is empty, there's nothing left to subtract.
+		if len(final) == 0 {
+			return final
+		}
+
 		for k := range m {
 			delete(final, k)
 		}
--- a/sets/maps.go
+++ b/sets/maps.go
@@ -53,7 +53,7 @@ func IntersectMapKeys[K comparable, V any](maps ...map[K]V) map[K]V {
 		// Short-circuit: we've already got the smallest possible intersection
 		// (empty set), so there's no point in continuing.
 		if len(final) == 0 {
-			return final
+			return make(map[K]V)
 		}
 
 		// This is us, ignore
@@ -75,9 +75,9 @@ func IntersectMapKeys[K comparable, V any](maps ...map[K]V) map[K]V {
 }
 
 // UnionMapKeys finds the union of all m, where union is defined as the
-// combination of all keys in all m. In the case of conflict, the last-provided
-// key will be in the result. It always returns an allocated map, even if the
-// union is the empty set.
+// combination of all keys in all m. In the case where duplicate keys exist
+// across maps, the value corresponding to the key in the first map is used. It
+// always returns an allocated map, even if the union is the empty set.
 //
 // It does not modify any of the given inputs, but also does not deep copy any
 // values. That means the returned map may have keys and values that point to
@@ -97,7 +97,10 @@ func UnionMapKeys[K comparable, V any](maps ...map[K]V) map[K]V {
 	// Iterate over all maps and combine elements.
 	for _, m := range maps {
 		for k, v := range m {
-			final[k] = v
+			// Only overwrite if we haven't previously seen the value.
+			if _, ok := final[k]; !ok {
+				final[k] = v
+			}
 		}
 	}
 
@@ -126,7 +129,7 @@ func SubtractMapKeys[K comparable, V any](maps ...map[K]V) map[K]V {
 	for _, m := range maps[1:] {
 		// Short-circuit: if the map is empty, there's nothing left to subtract.
 		if len(final) == 0 {
-			return final
+			return make(map[K]V)
 		}
 
 		for k := range m {
--- a/sets/maps.go
+++ b/sets/maps.go
@@ -15,8 +15,9 @@
 package sets
 
 // IntersectMapKeys finds the intersection of all m keys, where intersection is
-// defined as keys that exist in all m. It always returns an allocated map, even
-// if the intersection is the empty set.
+// defined as keys that exist in all m. In the case where duplicate keys exist
+// across maps, the value corresponding to the key in the first map is used. It
+// always returns an allocated map, even if the intersection is the empty set.
 //
 // It does not modify any of the given inputs, but also does not deep copy any
 // values. That means the returned map may have keys and values that point to
--- a/sets/maps.go
+++ b/sets/maps.go
@@ -1,0 +1,132 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sets
+
+// IntersectMapKeys finds the intersection of all m keys, where intersection is
+// defined as keys that exist in all m. It always returns an allocated map, even
+// if the intersection is the empty set.
+//
+// It does not modify any of the given inputs, but also does not deep copy any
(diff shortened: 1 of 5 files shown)


sets/maps.go
```diff
@@ -0,0 +1,137 @@
...
+		final[k] = v
+	}
+
+	// Compute the intersection.
+	for i, m := range maps {
+		// Short-circuit: we've already got the smallest possible intersection
+		// (empty set), so there's no point in continuing.
+		if len(final) == 0 {
+			return final
+		}
```
verbanicm, line 56: nit: should this be moved outside the loop, it feels strange to be in here when you dont need to come into the loop for this.

sets/maps.go
```diff
@@ -0,0 +1,137 @@
...
+//
+// It does not modify any of the given inputs, but also does not deep copy any
+// values. That means the returned map may have keys and values that point to
+// the same objects as in the original map.
+func IntersectMapKeys[K comparable, V any](maps ...map[K]V) map[K]V {
+	if len(maps) == 0 {
+		return make(map[K]V)
+	}
+
+	// Pre-compute the maximum possible allocation to minimize allocs. Here we
```
pdewilde, line 30: This seems like it could way over-allocate depending on the actual data being input.

Wonder how it benchmarks vs for example a naieve approach or a two-pass using a bloom filter.

sets/maps.go
```diff
@@ -0,0 +1,137 @@
...
+	for k, v := range smallestMap {
+		final[k] = v
+	}
+
+	// Compute the intersection.
+	for i, m := range maps {
+		// Short-circuit: we've already got the smallest possible intersection
+		// (empty set), so there's no point in continuing.
+		if len(final) == 0 {
+			return final
```
pdewilde, line 56: I would 
`return make(map[K]V)`

to free up all the extra memory for the GC

sets/maps.go
```diff
@@ -0,0 +1,137 @@
...
+	initialMap := maps[0]
+	final := make(map[K]V, len(initialMap))
+	for k, v := range initialMap {
+		final[k] = v
+	}
+
+	for _, m := range maps[1:] {
+		// Short-circuit: if the map is empty, there's nothing left to subtract.
+		if len(final) == 0 {
+			return final
```
pdewilde, line 129: I would
`return make(map[K]V)`

to free up all the extra memory for the GC

sets/maps.go
sethvargo: We want to check on each iteration because lower on ~L68, we're deleting entries from the map that don't exist in the slice. So then when we go to process the next slice, we want to check again if the intersection is empty. We're modifying `final` in place.

sets/maps.go
```diff
@@ -0,0 +1,137 @@
...
+package sets
+
+// IntersectMapKeys finds the intersection of all m keys, where intersection is
+// defined as keys that exist in all m. It always returns an allocated map, even
+// if the intersection is the empty set.
+//
+// It does not modify any of the given inputs, but also does not deep copy any
+// values. That means the returned map may have keys and values that point to
+// the same objects as in the original map.
+func IntersectMapKeys[K comparable, V any](maps ...map[K]V) map[K]V {
```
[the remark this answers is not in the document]
sethvargo: I added more docs. Note that they will be the same _type_, but you're right that we could have:

```go
IntersectMapKeys(
  map[string]string{"a", "b"}
  map[string]string{"a", "c"}
)
```

in which case the value from the first entry takes precedence (returns `a:b`). There are tests for this, but I updated the docs. Thanks.

sets/maps.go
verbanicm: ack

sets/maps.go
sethvargo: I think the overhead of constructing a bloom filter would be fairly high (plus complexity). What do you mean by "naive approach"?

pdewilde: Don't preallocate and allow the runtime to grow the slice for you (should be same amortized runtime, worst case is 2x the memory pressure)

pdewilde: or log_2(length) the allocs

sets/slices.go
```diff
@@ -0,0 +1,196 @@
...
+// It does not modify any of the given slices, but also does not deep copy any
+// values. That means the returned slice may have elements that point to the
+// same objects as in the original slice.
+func Subtract[T comparable](slices ...[]T) []T {
+	if len(slices) == 0 {
+		return []T{}
+	}
+
+	// Make a copy of the first slice; we're going to gradually remove elements.
+	final := append([]T{}, slices[0]...)
```
pdewilde, line 157: Would it be faster/cleaner to not do the first call to append?

First generate the toDelete map from `slices[1:]` and then iterate once through `slices[0]` appending to `final` only if `slices[0][i]` is not in `toDelete`

pdewilde: Maybe rename `toDelete` to `toSkip`, `toSubtract` or `toOmit`

sets/slices.go
```diff
@@ -0,0 +1,196 @@
...
+	final := make([]T, 0, len(result))
+	for k := range result {
+		final = append(final, k)
+	}
+	return final
+}
+
+// IntersectStable has the same invariants [Intersect], except it preserves the
+// relative order of elements in the intersection. This function has a
+// comparatively higher runtime complexity than [Intersect].
```
pdewilde, line 56: Runtime could be maintained by just first running `IntersectMapKeys()` and then doing a linear pass through `slices` removing any key from the intersect map once seeing it.

pdewilde: Psudedocode:

```
maps  = []
for slice in slices:
   maps.append(sliceToSet(slice))
intersectSet := IntersectMapKeys(maps)

final = [len(intersectSet)]
for slice in slices:
   for element in slice:
       if element in intersectSet:
             intersectSet.delete(element)
             final.append(element)

```

sets/slices.go
```diff
@@ -0,0 +1,196 @@
...
+//
+// It does not modify any of the given slices, but also does not deep copy any
+// values. That means the returned slice may have elements that point to the
+// same objects as in the original slice.
+func Intersect[T comparable](slices ...[]T) []T {
+	if len(slices) == 0 {
+		return []T{}
+	}
+
+	maps := make([]map[T]struct{}, len(slices))
```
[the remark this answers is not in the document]
pdewilde: I guess its not that crazy though, just alloc per input map, and not a great way to get around it.

sets/maps.go
pdewilde: The more I think about it, the happier I am with how it is.

If you were really looking to go above and beyond, you could copy into a smaller array at the end before returning if memory savings would be large (maybe if over 10,000 elements and less than half full), but the user could do that themselves so its not a big deal.

sethvargo: Yea - I do that for the slices because it's pretty easy to trim off remaining capacity. It's less exciting for maps.

sets/slices.go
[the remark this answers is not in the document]
sethvargo: Yea - I couldn't figure out a good way. I'm open to ideas though.

sets/slices.go
sethvargo: I think it'll depend on the slice of the maps. For small maps (hundreds of elements), there's very little difference. Once you get into thousands or tens of thousands, this is comparatively more expensive. However, I would also argue that you shouldn't actually care about the order of 10k+ elements in a single slice.

sets/slices.go
```diff
@@ -58,24 +58,16 @@ func IntersectStable[T comparable](slices ...[]T) []T {
...
-	// Remove duplicates
-	finalSeen := make(map[T]struct{}, len(final))
-	var i int
-	for _, v := range final {
-		if _, ok := finalSeen[v]; !ok {
-			finalSeen[v] = struct{}{}
-			final[i] = v
-			i++
+	// Make a copy of the slice with duplicates removed.
+	final := make([]T, 0, len(slices[0]))
```
pdewilde, line 62: If you defer allocation until after finding uniq, you can size the slice to the exact size you need

sethvargo: Wouldn't that require another full iteration over `uniq`?

pdewilde: You're right, it would

pdewilde (approved):

yolocs (approved):

PR #167, opened 2023-07-23
lint: use our own terraform linter action

sethvargo:
I think this works as intended.

### DIFF
--- a/.github/workflows/terraform-lint.yml
+++ b/.github/workflows/terraform-lint.yml
@@ -49,15 +49,10 @@ jobs:
           repository: 'abcxyz/pkg'
           path: 'abcxyz-pkg'
 
-      - id: 'setup-go'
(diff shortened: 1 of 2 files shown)


bradegler (approved):

.github/workflows/terraform-lint.yml
```diff
@@ -49,15 +49,10 @@ jobs:
           repository: 'abcxyz/pkg'
           path: 'abcxyz-pkg'
 
-      - id: 'setup-go'
-        uses: 'actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753' # ratchet:actions/setup-go@v4
+      - id: 'lint'
+        uses: './.github/actions/terraform-linter'
```
verbanicm, line 53: This needs to reference the action from `./abcxyz-pkg/` since its checked out in the step above.

verbanicm: https://github.com/abcxyz/pkg/pull/169

PR #168, opened 2023-07-23
cli: expand flag helpers and add AfterParse

sethvargo:
This adds new helpers to the flag and flagset components which mirror more of the stdlib flag functions including `Arg()` and `Lookup()`. It also adds helpers for `LookupEnv` and `GetEnv`.

In combination, these items enable a powerful new feature: `AfterParseFuncs. As their name implies, these functions are invoked after flags are parsed (but before the `Parse` function returns). This enables post-processing of flag values to support use cases like:

- Defaulting the value of flagA to flagB's value. Today this requires custom logic in `Run`, because setting flagB's default value to flagA's value is resolved at compile time, not parse time. With `AfterParse`, it's safe to set the value of flagB to flagA, and it will happen automatically when `Parse` is called.

- Setting the value of a flag to an argument. Perhaps for backwards-compatibility or other API guarantees, you want to default the value of a flag to a command-line argument. Today this requires custom logic in `Run`, but with `AfterParse`, it's possible to set these values automatically when `Parse` is called.


cli/flags.go
```diff
@@ -142,14 +148,44 @@ func (f *FlagSet) NewSection(name string) *FlagSection {
...
+func (f *FlagSet) Lookup(name string) *flag.Flag {
+	return f.flagSet.Lookup(name)
+}
+
 // Args implements flag.FlagSet#Parse.
 func (f *FlagSet) Parse(args []string) error {
-	return f.flagSet.Parse(args)
+	// Call the normal parse function first, so that Args and everything are
+	// properly set for any after functions.
+	merr := f.flagSet.Parse(args)
```
yolocs, line 182: I wonder if we should return err right away if `Parse` fails. `afterParseFuncs` likely depend on the fully parsed flags. Otherwise, the error from the `afterParseFuncs` might not even make sense.

yolocs (commented):
What about flags validation? Would it make sense to do that in an AfterFunc?

cli/flags.go
sethvargo: Hmm - this is a good point. However, some developers might want to return "all" CLI errors at once instead of forcing the user to go through a bunch of validations one-by-one.

I suppose the signature for afterFunc could include `parseErr err`, but it feels a bit weird.

sethvargo:
> What about flags validation? Would it make sense to do that in an AfterFunc?

I think it'd be reasonable to do light-weight validation in an AfterFunc (e.g. make sure a duration is positive). I was more envisioning a situation in which you wanted to test the state of a Command's internal fields post-parsing, but did not want to run the entire `Run` logic (which is where we usually do deferred defaulting or validation). It might be easier/better to encourage developers to have a `(c *Command) parseFlags()` function that they use in tests to verify that flags are parsed correctly.

cli/flags.go
sethvargo: Since we see use cases for both, I guess this question remains: is this functionality broadly useful, or would it make more sense to provide examples of how developers can do this themselves with a private function?

yolocs: Good question! I often find myself defining a struct to hold all flags (Dave also recent did this in abc CLI). I'm thinking an ideal experience like:

```go
set := cli.NewFlagSet()

type MyFlags struct {
  // The fields will be my flags
}

// Interface - minimally needed to register as a flag section.
func (f *MyFlags) Register(s *cli.FlagSet) {
  // Create a new section, register flags.
}

// This implements post parse hook interface.
func (f *MyFlags) AfterParse() error {
  // Post parse hook. Defaulting or validation.
}

f1 := &MyFlags{}

// Register a new section
// This calls the Register() above.
set.Register(f1)

// The parse calls AfterParse()
err := set.Parse(args)
```

sethvargo: Hmm - I don't love magical interfaces as an API construct. I think I prefer [how Guardian does it](https://github.com/abcxyz/guardian/blob/main/pkg/flags/github.go#L28) where each flagset is manually registered in the call to `Flags()`. Because of how flags get added, we also don't have access to `MyFlags` during parsing, so it'd need to be something like:

```go
type MyFlags struct {
  flagA string
  flagB string
}

func (f *MyFlags) Register(set *cli.FlagSet) {
  f := set.NewSection("MY FLAGS")

  f.StringVar(...)

  set.AfterFunc(func(args []string) error {
    if f.flagA == "" {
      return fmt.Errorf("-flag-a is required")
    }
  })
}
```

sethvargo: > "I don't love magical interfaces as an API construct"

I meant that I prefer to avoid situations in which a specially-named function changes the behavior of an upstream call. It's fine to implement an interface and then accept that type as an interface:

```go
type Fooer interface {
  Foo()
}

func DoSomething(f Fooer) {
  f.Foo()
}
```

but I dislike (and think it's a bad experience) when functions do something magic with interfaces (example):

```go
type Fooer interface {
  Foo()
}

func DoSomething(f Fooer) {
  if t, ok := f.(interface{
    Bar()
  }); ok {
    t.Bar()
  } else {
    t.Foo()
  }
}
```

because then suddenly defining a function on your struct can cause an unintended behavior change.

> That's also how JVS does it these days...

I think we've subtly converged on a similar pattern. However, I think there's some CUJs that remain unsolved in that pattern:

- If I have a separate flags struct, how and where do I provide validation for that structure? Should the validation live alongside my flagset or should it be in the parent command?

- If I need to do other post-processing (such as deferred defaulting), where does that happen? What if it needs to be reusable?

yolocs: > If I have a separate flags struct, how and where do I provide validation for that structure? Should the validation live alongside my flagset or should it be in the parent command?

I tend to define the validation alongside my flags struct. But there could be use cases where defining it in the command makes more sense. I think the approach in the PR provides the flexibility to do either.

> If I need to do other post-processing (such as deferred defaulting), where does that happen? What if it needs to be reusable?

With `AfterFunc`, I would default to have defaulting and validation logic defined as `AfterFunc` but will keep them alongside my flag struct. Like your example here: https://github.com/abcxyz/pkg/pull/168#discussion_r1274107967

That said, should we merge this PR?

---

Thanks for the explanation. I actually regard the interface magic as a Go strength :) Not your exact example, but something like:

```go
func DoSomething(f Fooer) {
  if t, ok := f.(interface{
    SetDefault()
  }); ok {
    // If the input has no defaulting logic,
    // then it doesn't have to implement an empty SetDefault func.
    t.SetDefault()
  }
  t.Foo()
}
```

sethvargo: > That said, should we merge this PR?

I think there's still the remaining question about whether we run `AfterFunc`s when normal flag parsing fails.

yolocs: > Since we see use cases for both, I guess this question remains: is this functionality broadly useful, or would it make more sense to provide examples of how developers can do this themselves with a private function?

I think it's useful. If I had this, I would have put the defaulting/validation logic there. Like your example:

```go
func (f *MyFlags) Register(set *cli.FlagSet) {
  f := set.NewSection("MY FLAGS")

  f.StringVar(...)

  set.AfterFunc(func(args []string) error {
    if f.flagA == "" {
      return fmt.Errorf("-flag-a is required")
    }
  })
}
```

sethvargo: Okay - I updated the code and the examples, plus added tests.

cli/flags.go
```diff
@@ -142,14 +148,52 @@ func (f *FlagSet) NewSection(name string) *FlagSection {
...
+
+	for _, fn := range f.afterParseFuncs {
+		func() {
+			defer func() {
```
yolocs, line 186: Fancy!

yolocs (approved):
We should update https://github.com/abcxyz/readability/wiki/Go-Best-Practices#command-line-tools-clis afterwards?

sethvargo:
> We should update https://github.com/abcxyz/readability/wiki/Go-Best-Practices#command-line-tools-clis afterwards?

Yup, will do shortly.

PR #172, opened 2023-07-27
feat: add bq and gcs helper

sailorlqh:
Add these helper functions per discussion: https://github.com/abcxyz/pmap/pull/159#discussion_r1276541648

I didn't include the retry for querying bq table as there will be different early stop functions and different retry configures base on different needs.


yolocs (commented):
/hold

yolocs:
I added retry and test in https://github.com/abcxyz/pkg/pull/173

testutil/bq.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+
+	"cloud.google.com/go/bigquery"
+	"github.com/sethvargo/go-retry"
+	"google.golang.org/api/iterator"
+)
+
+// QueryBQEntries queries the BQ and checks if a bigqury entry matched the query exists or not and return the results.
+// [T any] needs to be bigquery iteration supported type (e.g. string, bigquery.Value, struct{}).
+func QueryBQEntries[T any](ctx context.Context, query *bigquery.Query) ([]*T, error) {
```
pdewilde, line 30: So my understanding was [that interface types were always pointers,](https://jordanorelli.com/post/32665860244/how-to-use-interfaces-in-go) but seeing []*T being returned here seems to make me think that maybe that is not the case. 

Is it that on compilation time, any call to QueryBQEntries will have T be wrapped/unwrapped in an interface, so the []*T is still needed, as outside of the context of QueryBQEntries, T is still a concerte type?

testutil/bq.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return nil, retry.RetryableError(fmt.Errorf("failed to run query: %w", err))
+	}
+
+	if status, err := job.Wait(ctx); err != nil {
+		return nil, retry.RetryableError(fmt.Errorf("failed to wait for query: %w", err))
+	} else if err = status.Err(); err != nil {
+		return nil, retry.RetryableError(fmt.Errorf("query failed: %w", err))
+	}
+	it, err := job.Read(ctx)
```
pdewilde, line 41: Above lines could be combined using query.Read() rather than job.Read(), with the cost of less-specific error messages.

testutil/bq.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return nil, retry.RetryableError(fmt.Errorf("failed to read job: %w", err))
+	}
+
+	var entries []*T
+	for {
```
pdewilde, line 47: Not a huge fan of while loop with break, is it because you don't want to define `var entry T` outside of the for loop?

I do see it is what is used in the example code in the BQ docs, so it may just be me who has that particular nit :)

testutil/gcs.go
```diff
@@ -0,0 +1,46 @@
...
+// UploadGCSFiles uploads an object to GCS bucket.
+func UploadGCSFile(ctx context.Context, gcsClient *storage.Client, bucket, object string, data io.Reader, metadata map[string]string, conditions *storage.Conditions) error {
+	o := gcsClient.Bucket(bucket).Object(object)
+
+	// Set the upload conditions.
+	// See: https://pkg.go.dev/cloud.google.com/go/storage#Conditions
+	o = o.If(*conditions)
+
+	// Upload an object with storage.Writer.
+	wc := o.NewWriter(ctx)
```
pdewilde, line 35: [It is the caller's responsibility to call Close when writing is done. To stop writing without saving the data, cancel the context.] (https://pkg.go.dev/cloud.google.com/go/storage#ObjectHandle.NewWriter)

Should close/cancel be implemented?

pdewilde (approved):

sailorlqh:
Thanks @pdewilde for the comments, Chen made another [PR](https://github.com/abcxyz/pkg/pull/173) will more functionality and was merged earlier. This PR is now closed.

sailorlqh:
Thanks @pdewilde for the comments, Chen made another https://github.com/abcxyz/pkg/pull/173 will more functionality and was merged earlier. This PR is now closed.

PR #173, opened 2023-07-28
feat: add bqutil to retry query data in bigquery

yolocs:
cc @sailorlqh


bqutil/query.go
```diff
@@ -0,0 +1,136 @@
...
+	"cloud.google.com/go/bigquery"
+	"github.com/abcxyz/pkg/logging"
+	"github.com/sethvargo/go-retry"
+	"google.golang.org/api/iterator"
+)
+
+// Query is the interface to execute a query.
+type Query[T any] interface {
+	// Execute runs the query and returns the result in type []*T.
+	Execute(context.Context) ([]*T, error)
```
sethvargo, line 32: I'd probably make this `T` instead of `*T`. As written, if Query[*string], then you'd have to return a `**string` (which is technically valid, but super weird).

sethvargo: `NewQuery[*queryResult]()`

yolocs: You're right. Not sure what I was thinking.

sethvargo (approved):

PR #174, opened 2023-07-31
cli: fix string map variable flag only parsing the first KV pair

sJJdGG:


yolocs:
Hi @sJJdGG, thanks for you contribution. But I think the current behavior is intended. You are supposed to specify the flag multiple times. E.g. `cli -map-flag=key1=val1 -map-flag=key2=val2`

sethvargo:
Hi @sJJdGG - this is working as intended. If you want to set multiple values, use the flag multiple times. For example, if you're flag was:

```go
m := make(map[string]string])
f.StringMapVar(&cli.StringMapVar{
  // ...
  Name:   "foo",
  Target: &m,
})
```

Running `mycli -foo=a=b -foo=c=d` would result in the map having `{a:b, c:d}`. Is that not working?

We intentionally do not want to support multiple K=V pairs in the same flag because sometimes values have special characters like commas or semicolons, which makes finding a good delimiter a challenge.

sethvargo:
Thanks - yea we need to write some tests for the various CLI flags, but we're still rounding out some of the APIs.

The printer does print out _all_ values, and that's intentional.

PR #175, opened 2023-08-01
Support back-and-forth dialogs in Prompt() tests

drevell:
Why? In the abc CLI we want to test a function that does multiple rounds of back-and-forth dialog with the user. Currently, this isn't possible, because prompts don't appear in a test environment. `Prompt()`
doesn't write to stdout when stdout is a `*bytes.Buffer` (as created by `BaseCommand.Pipe()`).

How do we do this? We add a type-assertion to check is stdout is a `*io.PipeWriter.` If so, we assume that it's appropriate to print a prompt. Tests that want to see prompt messages can just use `io.Pipe`.

Also, we add  a new test`TestBaseCommand_Prompt_Dialog`, which demonstrates how to write a kind of "conversational" tests using `Prompt()`.

*Alternative considered*: rather than checking for stdout to be a `*io.PipeWriter`, check that it implements some special interface like `type promptable interface { promptable() }`. Callers would then have to wrap their `io.Writer` in a type that implements that interface to indicate their willingness to receive promopts. This would be more flexible, but also more verbose and harder to understand.

### DIFF
--- a/cli/command.go
+++ b/cli/command.go
@@ -280,7 +280,7 @@ func (c *BaseCommand) Hidden() bool {
 // up to 64k bytes. If there's an input stream (e.g. a pipe), it will read the
 // pipe.
 //
-// The prompt will be printed to stdout if any of these cases is true:
+// The prompt will be printed to c.Stdout() if any of these cases is true:
 //   - the terminal is a TTY (for real user interaction)
 //   - c.StdIn(), c.Stdout(), and c.Stderr() came from io.Pipe() (for unit-testing back-and-forth dialog)
 //
--- a/cli/command.go
+++ b/cli/command.go
@@ -287,7 +287,7 @@ func (c *BaseCommand) Hidden() bool {
 // It will fail if stdin pipe and the terminal is not a tty. If the context is canceled,
 // this function leaves the c.Stdin in a bad state.
 func (c *BaseCommand) Prompt(ctx context.Context, msg string, args ...any) (string, error) {
-	_, stdoutIsPipe := c.Stdout().(*io.PipeWriter) // true if this is a unit test
+	_, stdoutIsPipe := c.Stdout().(*io.PipeWriter) // we expect unit tests to use PipeWriter if they want to see prompts
 
 	stdinIsTTY := c.Stdin() == os.Stdin && isatty.IsTerminal(os.Stdin.Fd())
 
--- a/cli/command.go
+++ b/cli/command.go
@@ -278,11 +278,20 @@ func (c *BaseCommand) Hidden() bool {
 
 // Prompt provides a mechanism for asking for user input. It reads from [Stdin],
 // up to 64k bytes. If there's an input stream (e.g. a pipe), it will read the
-// pipe. If the terminal is a TTY, it will prompt. Otherwise it will fail if
-// there's no pipe and the terminal is not a tty. If the context is canceled,
+// pipe.
+//
+// The prompt will be printed to stdout if any of these cases is true:
+//   - the terminal is a TTY (for real user interaction)
+//   - c.Stdout() is an *io.PipeReader (for unit-testing back-and-forth dialog)
+//
+// It will fail if stdin pipe and the terminal is not a tty. If the context is canceled,
 // this function leaves the c.Stdin in a bad state.
 func (c *BaseCommand) Prompt(ctx context.Context, msg string, args ...any) (string, error) {
-	if c.Stdin() == os.Stdin && isatty.IsTerminal(os.Stdin.Fd()) {
+	_, stdoutIsPipe := c.Stdout().(*io.PipeWriter) // true if this is a unit test
+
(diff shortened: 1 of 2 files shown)


cli/command_test.go
```diff
@@ -381,6 +383,117 @@ func TestBaseCommand_Prompt_Cancels(t *testing.T) {
...
+	}
+	if iceCream != "mint chip" {
+		t.Fatalf(`got iceCream %q, wanted "mint chip"`, iceCream)
+	}
+}
+
+// readWithTimeout does a single read from the given reader. It calls Fatal if
+// that read fails or the returned string doesn't contain wantSubStr. May leak a
+// goroutine on timeout.
+func readWithTimeout(t *testing.T, r io.Reader, wantSubstr string) {
```
sethvargo, line 442: Nit: prefer testing.TB interface instead (in case we ever needed to use this in benchmarks)

cli/command_test.go
```diff
@@ -381,6 +383,117 @@ func TestBaseCommand_Prompt_Cancels(t *testing.T) {
...
+	}
+
+	if !strings.Contains(got, wantSubstr) {
+		t.Fatalf("got a prompt %q, but wanted a prompt containing %q", got, wantSubstr)
+	}
+}
+
+// writeWithTimeout does a single write to the given writer. It calls Fatal
+// if that read dosen't contain wantSubStr. May leak a goroutine on timeout.
+func writeWithTimeout(t *testing.T, w io.Writer, msg string) {
```
sethvargo, line 476: Nit: prefer testing.TB interface instead (in case we ever needed to use this in benchmarks)

cli/command_test.go
```diff
@@ -381,6 +383,117 @@ func TestBaseCommand_Prompt_Cancels(t *testing.T) {
...
+	case <-time.After(time.Second):
+		t.Fatal("timed out waiting for background Prompt() goroutine to exit")
+	}
+
+	if color != "blue" {
+		t.Fatalf(`got color %q, wanted "blue"`, color)
+	}
+	if iceCream != "mint chip" {
+		t.Fatalf(`got iceCream %q, wanted "mint chip"`, iceCream)
+	}
```
sethvargo, line 436: Nit: prefer `got,want` style:

```suggestion
	if got, want := color, "blue"; got != want {
		t.Fatalf("expected color %q to be %q", got, want)
	}
	if got, want := iceCream, "mint chip"; got != want {
		t.Fatalf("expected iceCream %q to be %q", got, want)
	}
```

cli/command_test.go
```diff
@@ -381,6 +383,117 @@ func TestBaseCommand_Prompt_Cancels(t *testing.T) {
...
+// writeWithTimeout does a single write to the given writer. It calls Fatal
+// if that read dosen't contain wantSubStr. May leak a goroutine on timeout.
+func writeWithTimeout(t *testing.T, w io.Writer, msg string) {
+	t.Helper()
+
+	t.Logf("writeWithTimeout starting with %q", msg)
+
+	errCh := make(chan error)
+	go func() {
+		_, err := w.Write([]byte(msg))
```
sethvargo, line 483: `defer close(errCh)`

cli/command.go
```diff
@@ -278,11 +278,20 @@ func (c *BaseCommand) Hidden() bool {
 
 // Prompt provides a mechanism for asking for user input. It reads from [Stdin],
 // up to 64k bytes. If there's an input stream (e.g. a pipe), it will read the
-// pipe. If the terminal is a TTY, it will prompt. Otherwise it will fail if
-// there's no pipe and the terminal is not a tty. If the context is canceled,
+// pipe.
+//
+// The prompt will be printed to stdout if any of these cases is true:
```
sethvargo, line 283:
```suggestion
// The prompt will be printed to c.Stdout() if any of these cases is true:
```

cli/command.go
```diff
@@ -278,11 +278,20 @@ func (c *BaseCommand) Hidden() bool {
...
+//   - the terminal is a TTY (for real user interaction)
+//   - c.Stdout() is an *io.PipeReader (for unit-testing back-and-forth dialog)
+//
+// It will fail if stdin pipe and the terminal is not a tty. If the context is canceled,
 // this function leaves the c.Stdin in a bad state.
 func (c *BaseCommand) Prompt(ctx context.Context, msg string, args ...any) (string, error) {
-	if c.Stdin() == os.Stdin && isatty.IsTerminal(os.Stdin.Fd()) {
+	_, stdoutIsPipe := c.Stdout().(*io.PipeWriter) // we expect unit tests to use PipeWriter if they want to see prompts
```
sethvargo, line 290: I think this is fine, but:

- Should we check more than stdout? I'm wondering if we should check that stderr and stdout are pipewriters and stdin is a pipereader, just so we don't have a false positive hit.
- Should we move this into a helper function like `shouldPrintPrompt() bool`?

drevell: Both done.

sethvargo (approved):
The linter is mad, but this LGTM.

PR #178, opened 2023-08-05
cli: add PromptAll and PromptTo

sethvargo:
As identified in abcxyz/lumberjack#433, sometimes we need to read the entire value of stdin, until EOF. The current implementation of `Prompt` relies on newline characters for termination. The new implementation has a generic `PromptTo` function that leverages a `bufio.SplitFunc` to customize splitting. It also adds a `PromptAll` function that will read everything until EOF.

This PR also refactors the tests a bit to unify all the tests for a particular function under a single test (with everything else under subtests).

### DIFF
--- a/cli/command.go
+++ b/cli/command.go
@@ -282,6 +282,11 @@ func (c *BaseCommand) Hidden() bool {
 // For more information about the conditions under which the prompt is
 // displayed, see [PromptTo].
 func (c *BaseCommand) Prompt(ctx context.Context, msg string, args ...any) (string, error) {
+	// Unlike [bufio.ScanLines], this function reads at most a single line and
+	// then returns. bufio.ScanLines yields for each line, but this function
+	// yields a single line and then marks the read as terminated. This ensures we
+	// read up to the first newline character, but ignore lines thereafter
+	// (matching the original implementation's behavior).
 	splitFunc := func(data []byte, atEOF bool) (advance int, token []byte, err error) {
 		if atEOF && len(data) == 0 {
 			return 0, nil, nil
@@ -349,10 +354,8 @@ func (c *BaseCommand) PromptTo(ctx context.Context, splitFunc bufio.SplitFunc, m
 		scanner.Split(splitFunc)
 
 		for scanner.Scan() {
-			select {
-			case <-ctx.Done():
+			if ctx.Err() != nil {
 				return
-			default:
 			}
 
 			b.WriteString(scanner.Text())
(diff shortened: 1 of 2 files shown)


cli/command.go
```diff
@@ -276,38 +276,110 @@ func (c *BaseCommand) Hidden() bool {
...
+
+		scanner := bufio.NewScanner(stdin)
+		scanner.Split(splitFunc)
+
+		for scanner.Scan() {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
```
drevell, line 356: Optional: you can avoid the `select` by checking `ctx.Err() != nil`. Because ctx.Err() returns nil if any only if the ctx is done.

Same below.

cli/command_test.go
```diff
@@ -273,29 +274,112 @@ func TestRootCommand_Run(t *testing.T) {
...
-			args:   []string{"prompt"},
-			msg:    "enter input value:",
-			inputs: []string{"input"},
-		}, {
-			name:   "handles_multiple_prompts",
-			args:   []string{"prompt"},
-			msg:    "enter input value:",
-			inputs: []string{"input1", "input2"},
+			name: "reads_line",
+			execute: func(t *testing.T, cmd *BaseCommand) {
```
drevell, line 289: I'd suggest that it makes more sense for these to be separate test rather than defining these closures. There's not much benefit that I see to combining these three subtest into a single test.

drevell (changes requested):

cli/command.go
sethvargo: Good call, done.

cli/command.go
```diff
@@ -276,38 +276,110 @@ func (c *BaseCommand) Hidden() bool {
...
-// Prompt provides a mechanism for asking for user input. It reads from [Stdin],
-// up to 64k bytes. If there's an input stream (e.g. a pipe), it will read the
-// pipe.
+// Prompt asks for user input and reads it from [Stdin] until it encounters a
+// newline character. If there's an input stream (e.g. a pipe), it will read the
+// pipe. The result will not include the trailing newline or carriage return.
+// For more information about the conditions under which the prompt is
+// displayed, see [PromptTo].
+func (c *BaseCommand) Prompt(ctx context.Context, msg string, args ...any) (string, error) {
+	splitFunc := func(data []byte, atEOF bool) (advance int, token []byte, err error) {
```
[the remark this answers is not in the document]
sethvargo: Done.

cli/command_test.go
sethvargo: I've generally always followed the belief that all tests for a logical unit (in this case a function) should reside under `Test<Thing>`, and use subtests to group the logic and output into something reasonable. That way, when looking at test output, it's immediately clear which struct, function, and subtest is failing. I don't have to spelunk around multiple test functions to figure out where a failure is coming fromt.

I can break them apart into their own `t.Run()` statements if you think that'd be useful, but I don't want to create `TestPrompt_reads_line`, `TestPrompt_reads_multiple`, etc. I find that extremely difficult to parse.

The disadvantage of moving away from the subtest is that we have to duplicate the cli setup logic. If that were ever refactored, it would need to be updated in many places instead of just once.

drevell: OK, I'm fine with that.

drevell (approved):

PR #180, opened 2023-08-09
Re-disable caching in Go CI

sethvargo:
This appears to be breaking golangci and a few other things. We don't get a ton of benefit from these caches.


sethvargo:
Hmm maybe we do get a benefit; these are much slower. I guess its because we're also caching the build, not just the deps.

PR #184, opened 2023-08-16
Avoid overflowing int and uint on 32-bit systems

sethvargo:
This fixes a bug where, on 32-bit operating systems, a large integer value would overflow when boxed from a 64-bit int/uint (from ParseInt/ParseUint) to a 32bit int/uint.


sethvargo:
- https://github.com/abcxyz/pkg/security/code-scanning/3
- https://github.com/abcxyz/pkg/security/code-scanning/2

shan4993 (approved):

sailorlqh (approved):

PR #188, opened 2023-08-17
TestLogger logs at most specific level with sources

sethvargo:
Also apply our custom mutator so the fields match.

### DIFF
--- a/logging/testing.go
+++ b/logging/testing.go
@@ -17,6 +17,7 @@ package logging
 import (
 	"io"
 	"log/slog"
+	"math"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


logging/testing.go
```diff
@@ -26,17 +27,20 @@ func TestLogger(tb testing.TB) *slog.Logger {
 	tb.Helper()
 
 	w := &testingWriter{tb}
-	return slog.New(slog.NewTextHandler(w, &slog.HandlerOptions{
+	level := slog.Level(math.MinInt)
```
drevell, line 30: Could you add a comment here? This reasoning behind this isn't immediately obvious to me.

sethvargo: Done.

drevell (approved):

PR #189, opened 2023-08-17
Also accept "WARN" and "ERR" as valid log level names

sethvargo:


logging/levels.go
```diff
@@ -77,9 +77,9 @@ func LookupLevel(name string) (slog.Level, error) {
 		return LevelInfo, nil
 	case levelNoticeName:
 		return LevelNotice, nil
-	case levelWarningName:
+	case levelWarningName, "WARN":
```
drevell, line 80: Should the same pattern be applied to the other levels? Is there a reason for these two to behave differently?

drevell (changes requested):

logging/levels.go
sethvargo: This is to maintain backwards compat with the old loggers levels. The other levels remain the same. I can document that if desired.

drevell: Yeah, might be worth a note. I don't have strong feelings on it.

drevell (approved):

logging/levels.go
sethvargo: Added docs.

drevell (approved):

Issue #211, opened 2023-09-14
[cli][flags] StringSliceVar Default value is appended even if user passes their own values


dcreey:
### TL;DR

The default value for a StringSliceVar is always appended to the result even if the user passes their own values.

Example:
```
        f.StringSliceVar(&cli.StringSliceVar{
		Name:    "github-issue-labels",
		Target:  &d.FlagGitHubIssueLabels,
		Example: "guardian-iam-drift",
		Usage:   `The labels to use on any created GitHub Issues.`,
		Default: []string{"guardian-statefile-drift"},
	})
```

if the user passes `-github-issue-labels="guardian-statefile-drift,security"` the `FlagGitHubIssueLabels` value will have 3 values present: `guardian-statefile-drift guardian-statefile-drift security`

### Expected behavior

The default value is only present if the user has not passed their own values for this flag.

### Observed behavior

The default values(s) are always present even if the user passes their own.

### Environment Details

```markdown
Ubuntu-latest on GitHub action. Go 1.21.
```

PR #212, opened 2023-09-14
Remove defaults when a value is given in maps and slices

sethvargo:
Fixes GH-211

### DIFF
--- a/cli/flags_test.go
+++ b/cli/flags_test.go
@@ -218,6 +218,51 @@ func TestFlagSection_StringMapVar(t *testing.T) {
 				"c": "d",
 			},
 		},
+		{
+			name: "given_default_value_one",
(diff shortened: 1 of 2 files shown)


yolocs (approved):

cli/flags_test.go
```diff
@@ -160,6 +160,161 @@ func TestFlagSet_Default(t *testing.T) {
 	})
 }
 
+func TestFlagSection_StringMapVar(t *testing.T) {
```
dcreey, line 163: Can you add some test cases for when someone passes a value equal to the default as the first arg and they have more than one arg? Just to confirm that it won't remove the user's first arg

dcreey (dismissed):

yolocs (approved):

dcreey (approved):

PR #214, opened 2023-09-29
Add reusable create tag workflow

sqin2019:


.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,118 @@
...
+              }
+            }
+      - name: 'Create Tag'
+        uses: 'actions/github-script@98814c53be79b1d30f795b907e553d8679345975' # ratchet:actions/github-script@v6
+        with:
+          github-token: '${{ steps.mint-token.outputs.token }}'
+          retries: '3'
+          script: |+
+            let sha = '${{ github.sha }}'
+            if ('${{ inputs.branch }}' !== 'main') {
```
verbanicm, line 85: I believe there is a script injection here. Dont trust anything from the users. Instead put this in an env var and use process.env.MY_VAR instead so it is escape. With the currently syntax anything can be rendered in there.

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,118 @@
...
+              // Get branch.
+              const branch = await github.rest.repos.getBranch({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                branch: '${{ inputs.branch }}',
```
verbanicm, line 90: Same here

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,118 @@
...
+              })
+              // Branch returns the latest commit info. See schema:
+              // https://docs.github.com/en/rest/branches/branches?apiVersion=2022-11-28#get-a-branch
+              sha = branch['data']['commit']['sha']
+            }
+
+            await github.rest.git.createTag({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              tag: '${{ inputs.tag }}',
```
verbanicm, line 100: Same here

.github/workflows/create-tag.yml
sqin2019: Noted, thanks!

verbanicm, line 85: We should instead check against the default branch instead of main --> `${{ github.event.repository.default_branch }}`

.github/workflows/create-tag.yml
verbanicm: Validate these throughout

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,118 @@
...
+              object: sha,
+              type: 'commit',
+              message: '${{ inputs.message }}'
+            });
+
+            if ('${{ inputs.annotated_tag }}' === 'true') {
```
verbanicm, line 106: nit: we can check for "truthy" values like


```yaml
      if: |
        contains(fromJSON('["true", "True", "TRUE", "1", "T", "t"]'), inputs.annotated_tag)
```

verbanicm: disregard, didnt realize this was inside github script

.github/workflows/create-tag.yml
sqin2019: on a second thought, we should check against `[github.ref_name](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)` instead. Here is to get the sha of the requested branch (to create tags on). if the requested branch is the same branch as the event ref_name, I can just use `github.sha`

verbanicm: i dont think that is the case, we want the workflow to be run from main right, but the target branch might be something else

sqin2019: I could be missing something, but my understanding is that the workflow branch != input branch. If you are talking about the branch where the workflow is from, We can use deployment protected/selected branches to ensure it is allowlisted by us.
Screenshot 2023-09-29 at 1 53 05 PM

sqin2019: I tried to use workflow from a different branch (tmp) [here](https://github.com/sqin2019/hello-world/actions/runs/6357165921/job/17267936647), `github.ref_name == workflow branch (tmp here)`.  Do you mean sth else with target branch?

verbanicm: I think what im saying is if i create a workflow on a random branch and then say hey run this workflow from my branch, its not reviewed. So we should make sure the workflow is run from main (or the default branch), always. The input branch is the branch in which we create the tag on, this does not necessarily mean that the workflow branch and the input branch are equal. We can chat monday if needed.

sqin2019: I see, we should really add a condition on the create-tag job, so we only run the job when `${{ github.ref_name }} === ${{ github.event.repository.default_branch }}

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,130 @@
...
+        description: 'URL for the production github-token-minter service.'
+        type: 'string'
+        required: true
+
+jobs:
+  create-tag:
+    runs-on: 'ubuntu-latest'
+    # Only running the job when the triggering branch is the default branch.
+    if: '${{ github.ref_name == github.event.repository.default_branch }}'
+    environment: '${{ inputs.deployment_environment }}'
```
verbanicm, line 63: Is there a possibility to give a non existent environment and this will bypass the approval somehow?

sqin2019: Assume if the workflow is from default_branch, there shouldn't be any concern. If we don't want to entirely rely on code review, then we need to do some validation here. also, I doubt the the job will proceed if the environment does not exist (will double check).

verbanicm: I don't think this should block, let's do our due diligence and see if we need to secure it. Thanks!

verbanicm (dismissed):

.github/workflows/create-tag.yml
sqin2019: Sadly it did go through https://github.com/sqin2019/hello-world/blob/main/.github/workflows/create_tag.yml#L47. It is hard coded in the workflow, so will rely on code reviews. 

Thanks! Happy weekend!

.github/workflows/create-tag.yml
yolocs: Why should we allow non default_branch?

sqin2019: If you mean why allow tags on non default_branch, as Mike mentioned we might have a dedicated branch for release which is not necessarily the default branch.

.github/workflows/create-tag.yml
yolocs: Should the environment live in the parent workflow? I'm not sure why this workflow cares about the env

.github/workflows/create-tag.yml
yolocs: I actually think it's a better idea to only allow this input once we really have a release branch. For one, our immediate goal is to avoid creating tags on non default branch (which is our only release branch). Having it as an input before we can enforce multi-party review is problematic. For two, we haven't discussed how release branches should work so introducing this input is a bit premature.

.github/workflows/create-tag.yml
sqin2019: We cannot use environment in the parent workflow, [here]( https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow) is how we can use environment in reusable workflow, I will add a comment for it, thx!

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,130 @@
...
+            {
+              "repositories": ["${{ github.event.repository.name }}"],
+              "permissions": {
+                "contents": "write"
+              }
+            }
+      - name: 'Create Tag'
+        uses: 'actions/github-script@98814c53be79b1d30f795b907e553d8679345975' # ratchet:actions/github-script@v6
+        env:
+          BRANCH: '${{ inputs.branch }}'
```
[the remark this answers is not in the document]
sqin2019: The triggering branch is the where the calling workflow is from, the workflow can create tag on different branches depending on our workflow logic.

.github/workflows/create-tag.yml
sqin2019: I see, for the interim, we can skip the job when the input branch is not default branch. updating, thx!

.github/workflows/create-tag.yml
yolocs: A more canonical reference for why it's not possible https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow

sqin2019: Added comment, thank you!

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,149 @@
...
+          result-encoding: 'string'
+          script: |+
+            const branch = process.env.BRANCH
+            const tag = process.env.TAG
+            const annotatedTag = process.env.ANNOTATED_TAG
+            const message = process.env.MESSAGE
+
+            // TODO(#215): Once prevent self-review in supported by GitHub, we can remove this condition.
+            if (branch != '${{ github.event.repository.default_branch }}') {
+              return `Skipped creating tag as currently we only support creating tags on repository default branch`
```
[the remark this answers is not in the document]
sqin2019: updated.

yolocs (approved):

verbanicm (approved):

PR #220, opened 2023-10-05
fix: new design for load workflow variables

verbanicm:
This changes to a single YAML config file instead of an env file per directory. Using the new approach, the config file can be locked down via CODEOWNERS to prevent tampering while additionally allowing CODEOWNERS entries to allow sub-teams to manage their own directories.

```text
env.yml                        owner/admin-team
team/project/environment       owner/project-team
```

Additionally, this new approach allows a * suffix to allow any sub directory to inherit the high level variables.

### DIFF
--- a/.github/actions/load-workflow-variables/action.yml
+++ b/.github/actions/load-workflow-variables/action.yml
@@ -18,93 +18,92 @@ description: |-
 
 inputs:
   working_directory:
-    description: 'The working directory to look for the env file.'
+    description: 'The working directory to look for the env file, relative to the GitHub workspace.'
     required: false
     default: '.'
-  env_filename:
-    description: 'The YAML filename containing variables to override with.'
+  filepath:
+    description: 'The YAML filepath to load the configuration from, relative to the GitHub workspace.'
     required: false
     default: 'env.yml'
-  find_in_parents:
-    description: 'If true, search for env_filename in parents until one is found.'
+  fail_on_missing:
+    description: 'If true, the process will exit with a non-zero exit code indicating failure when a value is not found.'
     required: false
-    default: 'false'
-  top_level_directory:
-    description: 'The top level directory to stop in when inheriting environment files.'
-    required: false
-    default: '.'
+    default: 'true'
 
 runs:
   using: 'composite'
   steps:
-    - name: 'Load From Local File Only'
-      if: |-
-        !contains(fromJSON('["true", "True", "TRUE", "1", "T", "t"]'), inputs.find_in_parents)
+    - name: 'Load Workflow Variables'
       shell: 'bash'
       env:
         WORKING_DIRECTORY: '${{ inputs.working_directory }}'
-        ENV_FILENAME: '${{ inputs.env_filename }}'
-        TOP_LEVEL_DIRECTORY: '${{ inputs.top_level_directory || github.workspace }}'
+        FILEPATH: '${{ inputs.filepath }}'
+        FAIL_ON_MISSING: '${{ inputs.fail_on_missing }}'
       run: |-
-        WORKING_DIRECTORY=$(realpath "${WORKING_DIRECTORY}")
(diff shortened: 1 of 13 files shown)


bradegler (approved):

.github/actions/load-workflow-variables/test/test.env.yml
```diff
@@ -0,0 +1,25 @@
+- dirs:
```
dcreey, line 1: What if we have 2 or more conflicting entries? What if we have a `*` on the root directory and then we have a lower level exact match?

.github/actions/load-workflow-variables/action.yml
```diff
@@ -18,93 +18,92 @@ description: |-
...
-              echo "FOUND"
-              break
-            fi
+                if [[ ! -z "${FOUND}" ]]; then
+                    echo "FOUND       Found match for ${SEARCH_DIR}"
+                    break
+                fi
 
-            echo "MISSING"
+                echo "NOT FOUND   No match for ${SEARCH_DIR}"
```
dcreey, line 86: Won't this print out on every iteration (even if it finds a match)?

.github/actions/load-workflow-variables/test/test.env.yml
verbanicm: Currently, the lower level would win and short circuit.

.github/actions/load-workflow-variables/action.yml
verbanicm: Theres a break statement in the if above

.github/actions/load-workflow-variables/action.yml
```diff
@@ -18,93 +18,92 @@ description: |-
...
-          echo "MISSING"
+        if [[ -z "${FOUND}" ]]; then
+            echo "NOT FOUND   No exact match for ${WORKING_DIRECTORY}"
 
-          # failsafe to prevent infinite loop
-          while [[ "${WORKING_DIRECTORY}" != "/"  ]]; do
-            WORKING_DIRECTORY=$(dirname "${WORKING_DIRECTORY}")
-            SEARCH_PATH="${WORKING_DIRECTORY}/${ENV_FILENAME}"
+            # Search directory hierarchy upwards for a * suffix indicating all children of parent be used
+            while [[ "${WORKING_DIRECTORY}" != "/" && "${WORKING_DIRECTORY}" != "." && "${WORKING_DIRECTORY}" != "" ]]; do
```
dcreey, line 76: This is pretty inefficient. What happens when the env file blows up (in the case of X's it will eventually get pretty large). It would probably be easier to just inline a python file and write that to tmp then execute it.

.github/actions/load-workflow-variables/action.yml
```diff
@@ -18,93 +18,92 @@ description: |-
...
       env:
         WORKING_DIRECTORY: '${{ inputs.working_directory }}'
-        ENV_FILENAME: '${{ inputs.env_filename }}'
-        TOP_LEVEL_DIRECTORY: '${{ inputs.top_level_directory || github.workspace }}'
+        FILEPATH: '${{ inputs.filepath }}'
+        FAIL_ON_MISSING: '${{ inputs.fail_on_missing }}'
       run: |-
-        WORKING_DIRECTORY=$(realpath "${WORKING_DIRECTORY}")
-        SEARCH_PATH="${WORKING_DIRECTORY}/${ENV_FILENAME}"
+        FULL_WORKING_DIR=$(realpath "${WORKING_DIRECTORY}")
```
dcreey, line 43: Overall I like it.

.github/actions/load-workflow-variables/action.yml
dcreey: break is only hit on the iteration where it matches the working dir. If you had to navigate up 4 times first it would print this message 4 times, no?

verbanicm: Yea its just informing the user of the search hierarchy.

PR #221, opened 2023-10-09
fix: paths might not exist when using load env vars

verbanicm:


bradegler (approved):

.github/workflows/load-workflow-variables-test.yml
```diff
@@ -4,7 +4,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#     http://www.apache.org/licenses/LICENSE-2.0
+#     http:/www.apache.org/licenses/LICENSE-2.0
```
pdewilde, line 7: Think you lost a `/`

.github/workflows/load-workflow-variables-test.yml
```diff
@@ -29,22 +29,22 @@ jobs:
       - name: 'Load Env Vars'
         uses: './.github/actions/load-workflow-variables' # ratchet:exclude
         with:
-          working_directory: '.github/actions/load-workflow-variables/test/terraform/org'
-          filepath: '${{ github.workspace }}/.github/actions/load-workflow-variables/test/test.env.yml'
+          working_directory: 'terraform/org'
+          filepath: '.github/actions/load-workflow-variables/test.env.yml'
```
pdewilde, line 33: Why did this filepath change?
```shell
OLD:  .github/actions/load-workflow-variables/test/test.env.yml
New:  .github/actions/load-workflow-variables/test.env.yml
```

pdewilde: I see now, file was moved. GH UI I thought usually detected that as a move rather than an addition and deletion.

PR #222, opened 2023-10-09
fix: license header

verbanicm:


pdewilde (approved):
Do you think this could be broken into smaller PRs? :grin:

PR #224, opened 2023-10-23
fix: allow empty token request property

verbanicm:
Allow repositories to be empty to allow tokens to be created for all granted repositories.


bradegler (approved):

sethvargo:
Hmm - I'm a little worried about this, since this is the default if you pass in no value or an empty value. I'd rather have a sentinel value (e.g. `*` or `$ALL_REPOS`) that must be explicitly provided. This isn't secure-by-default, since it's now impossible to distinguish the intention between:

```json
{
  "repositories":null
}
```

```json
{
  "repositories":[]
}
```

```json
{}
```

PR #227, opened 2023-10-24
fix: specific functions for github access tokens

verbanicm:
Create specific function for calling the GitHub API for generating installation access tokens.

### DIFF
--- a/githubapp/githubapp.go
+++ b/githubapp/githubapp.go
@@ -211,6 +211,10 @@ func (g *GitHubApp) generateAppJWT() ([]byte, error) {
 // access token for this application installation with the requested
 // permissions and repositories.
 func (g *GitHubApp) AccessToken(ctx context.Context, request *TokenRequest) (string, error) {
+	if request.Repositories == nil {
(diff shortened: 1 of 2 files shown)


bradegler (dismissed):

githubapp/githubapp.go
```diff
@@ -203,16 +211,35 @@ func (g *GitHubApp) generateAppJWT() ([]byte, error) {
 // access token for this application installation with the requested
 // permissions and repositories.
 func (g *GitHubApp) AccessToken(ctx context.Context, request *TokenRequest) (string, error) {
-	appJWT, err := g.AppToken()
```
sethvargo, line 206: This looks good, but should we return an error if request.Repositories is empty? We can say "use "AccessTokenAllRepos instead" or something.

sethvargo (approved):

githubapp/githubapp.go
verbanicm: I think yes, if nil throw error, if empty thats ok, meaning no permissions. I will update.

PR #229, opened 2023-10-26
Add instructions for using the create-tag.yml scheme

drevell:
And also autoformat for line length wrapping it seems

### DIFF
--- a/.github/workflows/create-tag.yml
+++ b/.github/workflows/create-tag.yml
@@ -12,6 +12,41 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Instructions for integrating with this workflow:
+#
+# Audience: you are an engineer working within the abcxyz GitHub org and
+# you want to call this workflow to create tags within your repo.
+#
+# Steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/create-tag.yml
```diff
@@ -12,6 +12,41 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Instructions for integrating with this workflow:
+#
+# Audience: you are an engineer working within the abcxyz GitHub org and
+# you want to call this workflow to create tags within your repo.
+#
+# Steps:
```
sqin2019, line 20: Thanks for doing this, documenting here definitely makes much more sense! We need another task to create the environment in order to enforce create-tag review might worth mentioning, see examples in https://docs.google.com/document/d/1edoDaTbJZd1_eGmMidDoGVimcO3WZZaRJ6j8rOp3VzQ/edit#heading=h.g2ns2mc3yah4.

sqin2019 (approved):

sqin2019 (approved):

PR #230, opened 2023-10-26
Add slice utility functions

lock14:
* Add generic `Map`, `Filter`, and `Reduce` functions for slices as I find myself wanting these functions frequently and they are not available in the standard library as far as I am aware.

### DIFF
--- a/slices/slices.go
+++ b/slices/slices.go
@@ -1,0 +1,62 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package slices
+
+// Map returns a slice consisting of the results of applying the given function
+// to the elements of the given slice.
+func Map[T1, T2 any](slice []T1, mapper func(T1) T2) []T2 {
+	if mapper == nil {
+		panic("nil mapping function provided")
+	}
+	if slice == nil {
+		return nil
+	}
+	result := make([]T2, len(slice))
+	for index, value := range slice {
+		result[index] = mapper(value)
+	}
+	return result
+}
+
+// Reduce performs a reduction on the elements of the given slice, using the
+// given initial value and an accumulator function, and returns the reduced value.
+func Reduce[T1, T2 any](slice []T1, initialValue T2, accumulator func(T2, T1) T2) T2 {
+	if accumulator == nil {
+		panic("nil accumulator function provided")
+	}
+	result := initialValue
+	for _, value := range slice {
+		result = accumulator(result, value)
+	}
+	return result
+}
+
+// Filter returns a slice consisting of the elements of the given slice that
+// match the given predicate.
+func Filter[T any](slice []T, predicate func(T) bool) []T {
+	if predicate == nil {
+		panic("nil predicate provided")
+	}
+	if slice == nil {
+		return nil
+	}
+	result := make([]T, 0)
+	for _, value := range slice {
+		if predicate(value) {
+			result = append(result, value)
+		}
+	}
+	return result
+}
--- a/slices/slices_test.go
+++ b/slices/slices_test.go
@@ -1,0 +1,226 @@
+// Copyright 2023 The Authors (see AUTHORS file)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package slices
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestMap(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		name   string
+		slice  []string
+		mapper func(string) int
+		want   []int
+	}{
+		{
+			name:   "nil_slice",
+			slice:  nil,
+			mapper: func(s string) int { return len(s) },
+			want:   nil,
+		},
+		{
+			name:   "empty_slice",
+			slice:  []string{},
+			mapper: func(s string) int { return len(s) },
+			want:   []int{},
+		},
+		{
+			name:   "one element",
+			slice:  []string{"foo"},
+			mapper: func(s string) int { return len(s) },
+			want:   []int{3},
+		},
+		{
+			name:   "many elements",
+			slice:  []string{"foo", "bar", "baz", "bing", "bang", "bong"},
+			mapper: func(s string) int { return len(s) },
+			want:   []int{3, 3, 3, 4, 4, 4},
+		},
+	}
+
+	for _, tc := range cases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			// Get a copy, so we can verify the input slices were not modified.
+			clone := testClone(tc.slice)
+
+			got := Map(tc.slice, tc.mapper)
+			if diff := cmp.Diff(tc.want, got); diff != "" {
+				t.Errorf("incorrect Map (-want/+got):\n%s", diff)
+			}
+
+			// Ensure we allocated up as much memory as possible.
+			if c, l := cap(got), len(got); c != l {
+				t.Errorf("expected cap(%d) to equal len(%d)", c, l)
+			}
+
+			// Ensure the slice was not modified.
+			if diff := cmp.Diff(clone, tc.slice); diff != "" {
+				t.Errorf("Map modified slice (-want/+got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestFilter(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		name      string
+		slice     []string
+		predicate func(string) bool
+		want      []string
+	}{
+		{
+			name:      "nil_slice",
+			slice:     nil,
+			predicate: func(s string) bool { return len(s) == 0 },
+			want:      nil,
+		},
+		{
+			name:      "empty_slice",
+			slice:     []string{},
+			predicate: func(s string) bool { return len(s) == 0 },
+			want:      []string{},
+		},
+		{
+			name:      "one_element_match",
+			slice:     []string{"foo"},
+			predicate: func(s string) bool { return len(s) == 3 },
+			want:      []string{"foo"},
+		},
+		{
+			name:      "one_element_no_match",
+			slice:     []string{"foo"},
+			predicate: func(s string) bool { return len(s) == 0 },
+			want:      []string{},
+		},
+		{
+			name:      "many_elements_all_match",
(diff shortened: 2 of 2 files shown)


stytchiz:
`map`, `filter`, and `reduce` are not idiomatic Go. The de-facto rule has been to use a for loop instead. Although these utility functions are easier to maintain with templates now, users can then write their Go code in a functional programming style which risks not being clean and concise. 

However, if there's much value to use these to reduce code duplication, then I won't object. I'll leave it to the other reviewers who have been here longer to decide that.

slices/slices.go
```diff
@@ -0,0 +1,62 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package slices
+
+// Map returns a slice consisting of the results of applying the given function
+// to the elements of the given slice.
+func Map[T1, T2 any](slice []T1, mapper func(T1) T2) []T2 {
+	if mapper == nil {
+		panic("nil mapping function provided")
```
stytchiz, line 21: It's hardly ever the case that code should panic: go/go-style/decisions#dont-panic. Returning an error and letting the caller handle it is more graceful.

lock14:
> `map`, `filter`, and `reduce` are not idiomatic Go. The de-facto rule has been to use a for loop instead. Although these utility functions are easier to maintain with templates now, users can then write their Go code in a functional programming style which risks not being clean and concise.
> 
> However, if there's much value to use these to reduce code duplication, then I won't object. I'll leave it to the other reviewers who have been here longer to decide that.

I'm not sure what isn't clean or concise about:
```
strings := []string{"a", "b", "c"}
lengths := slices.Map(strings, func(s string) int { return len(s) })
```

slices/slices_test.go
```diff
@@ -0,0 +1,226 @@
...
+		{
+			name:   "many elements",
+			slice:  []string{"foo", "bar", "baz", "bing", "bang", "bong"},
+			mapper: func(s string) int { return len(s) },
+			want:   []int{3, 3, 3, 4, 4, 4},
+		},
+	}
+
+	for _, tc := range cases {
+		tc := tc
```
stytchiz, line 58: As of Go 1.22, loop variable closures are not needed.

https://go.dev/blog/loopvar-preview

lock14: this library is still on 1.21

slices/slices.go
lock14: providing a `nil` mapping function is basically an egregious error. This is just attempting to provide a more informative message than the standard nil pointer deference. I don't think it worth polluting the API with an error worth it for this niche case.

slices/slices_test.go
```diff
@@ -0,0 +1,226 @@
...
+package slices
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestMap(t *testing.T) {
+	t.Parallel()
```
stytchiz, line 24: This Parallel call seems unnecessary if the subtests are being run in Parallel. I'm not sure what happens if Parallel is called twice in a goroutine/thread either.

lock14: see https://github.com/abcxyz/readability/wiki/Go-Best-Practices#table-driven

stytchiz:
> > `map`, `filter`, and `reduce` are not idiomatic Go. The de-facto rule has been to use a for loop instead. Although these utility functions are easier to maintain with templates now, users can then write their Go code in a functional programming style which risks not being clean and concise.
> > However, if there's much value to use these to reduce code duplication, then I won't object. I'll leave it to the other reviewers who have been here longer to decide that.
> 
> I'm not sure what isn't clean or concise about:
> 
> ```
> strings := []string{"a", "b", "c"}
> lengths := slices.Map(strings, func(s string) int { return len(s) })
> ```

Well, something even a little more complicated like ...

```
strings := []string{"a", "b", "c"}
lengths := slices.Map(strings, func(s string) int { return len(s) })
tot := slices.Reduce(lengths, func(n, m int) int {return n + m})  
```
vs.
```
strings := []string{"a", "b", "c"}
tot := 0
for _, s := range strings {
  tot += len(s)
}
```

but I'd be interested to hear what use cases in abcxyz would've been easier with this change.

lock14:
> > > `map`, `filter`, and `reduce` are not idiomatic Go. The de-facto rule has been to use a for loop instead. Although these utility functions are easier to maintain with templates now, users can then write their Go code in a functional programming style which risks not being clean and concise.
> > > However, if there's much value to use these to reduce code duplication, then I won't object. I'll leave it to the other reviewers who have been here longer to decide that.
> > 
> > 
> > I'm not sure what isn't clean or concise about:
> > ```
> > strings := []string{"a", "b", "c"}
> > lengths := slices.Map(strings, func(s string) int { return len(s) })
> > ```
> 
> Well, something even a little more complicated like ...
> 
> ```
> strings := []string{"a", "b", "c"}
> lengths := slices.Map(strings, func(s string) int { return len(s) })
> tot := slices.Reduce(lengths, func(n, m int) int {return n + m})  
> ```
> 
> vs.
> 
> ```
> strings := []string{"a", "b", "c"}
> tot := 0
> for _, s := range strings {
>   tot += len(s)
> }
> ```
> 
> but I'd be interested to hear what use cases in abcxyz would've been easier with this change.

I think It's hard to argue one is cleaner than the other as such a statement is very subjective.

stytchiz:
> > > > `map`, `filter`, and `reduce` are not idiomatic Go. The de-facto rule has been to use a for loop instead. Although these utility functions are easier to maintain with templates now, users can then write their Go code in a functional programming style which risks not being clean and concise.
> > > > However, if there's much value to use these to reduce code duplication, then I won't object. I'll leave it to the other reviewers who have been here longer to decide that.
> > > 
> > > 
> > > I'm not sure what isn't clean or concise about:
> > > ```
> > > strings := []string{"a", "b", "c"}
> > > lengths := slices.Map(strings, func(s string) int { return len(s) })
> > > ```
> > 
> > 
> > Well, something even a little more complicated like ...
> > ```
> > strings := []string{"a", "b", "c"}
> > lengths := slices.Map(strings, func(s string) int { return len(s) })
> > tot := slices.Reduce(lengths, func(n, m int) int {return n + m})  
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > vs.
> > ```
> > strings := []string{"a", "b", "c"}
> > tot := 0
> > for _, s := range strings {
> >   tot += len(s)
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but I'd be interested to hear what use cases in abcxyz would've been easier with this change.
> 
> I think It's hard to argue one is cleaner than the other as such a statement is very subjective.

Sure. There's no doubt in my mind that teams have written their own basic utility functions here just based on their subjective team-wide preferences. That's why I don't have a strong opinion.

From a readability standpoint though, I would call out go/go-style/guide#least-mechanism principle (prefer the standard way). That said, there's already a lot of discussion about this (e.g. https://yaqs.corp.google.com/eng/q/5103891194694336512#a1n6). With generics, there's definitely been some talk for the Go standard library owners to support such utility functions (perhaps in http://godoc/3/third_party/golang/go_exp/slices/slices).

lock14:
> > > > > `map`, `filter`, and `reduce` are not idiomatic Go. The de-facto rule has been to use a for loop instead. Although these utility functions are easier to maintain with templates now, users can then write their Go code in a functional programming style which risks not being clean and concise.
> > > > > However, if there's much value to use these to reduce code duplication, then I won't object. I'll leave it to the other reviewers who have been here longer to decide that.
> > > > 
> > > > 
> > > > I'm not sure what isn't clean or concise about:
> > > > ```
> > > > strings := []string{"a", "b", "c"}
> > > > lengths := slices.Map(strings, func(s string) int { return len(s) })
> > > > ```
> > > 
> > > 
> > > Well, something even a little more complicated like ...
> > > ```
> > > strings := []string{"a", "b", "c"}
> > > lengths := slices.Map(strings, func(s string) int { return len(s) })
> > > tot := slices.Reduce(lengths, func(n, m int) int {return n + m})  
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > vs.
> > > ```
> > > strings := []string{"a", "b", "c"}
> > > tot := 0
> > > for _, s := range strings {
> > >   tot += len(s)
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > but I'd be interested to hear what use cases in abcxyz would've been easier with this change.
> > 
> > 
> > I think It's hard to argue one is cleaner than the other as such a statement is very subjective.
> 
> Sure. There's no doubt in my mind that teams have written their own basic utility functions here just based on their subjective team-wide preferences. That's why I don't have a strong opinion.
> 
> From a readability standpoint though, I would call out the Go's principle of least mechanism (i.e. prefer the standard way). That said, there's already a lot of online discussion about this. With generics, there's definitely been some talk for the Go standard library owners to support such utility functions (perhaps in https://pkg.go.dev/golang.org/x/exp/slices).

That's fair, but I do think these functions will be added to the standard library at some point.  I'd rather not have to wait until they are added in order to use them. When/if they are added to the standard library it should be trivial to migrate to the standard library version as the function signatures will be very similar if not identical.

We did something similar with structured logging where we used a library for structured logging when the standard library didn't support it, and then migrated to the standard library version when it was released.

slices/slices_test.go
sethvargo: Brian is correct. Go 1.22 also isn't released yet. Once it's released, it will take 3-4 weeks for the ecosystem tooling to catch up. Then we can bump to 1.22 and consider dropping this. However, it will still work in 1.22 (it just isn't necessary), so we should continue doing this for now.

slices/slices_test.go
sethvargo: The first call to `t.Parallel()` causes `TestMap` to run in parallel with all the other tests marked `t.Parallel()`. The table-driven tests inside will run in serial to each other (but _potentially_ in parallel to other table-driven tests running in serial in a different top-level test).

Putting `t.Parallel()` on the subtests allows all subtests to run in parallel too. This effectively maximizes the number of tests running in parallel across all CPUs, and it's generally the way to get the fastest runtimes. This is especially true when the race detector is enabled.

sethvargo (changes requested):
Hey @lock14 - thanks for doing this PR. The tests look great and the style of your code is correct. However, I'm not supportive of this change, largely for the reasons @stytchiz has pointed out.

This is not idiomatic Go and we should aim to align as closely with the community as possible. There are plans to add higher-order operations on collections into the standard library (we got a few in 1.21, but more are supposed to be coming), at which point we can consider something like this.

In general, these kinds of list comprehensions only make sense on small data structures where the cost of allocation or storage is low. In those situations, the overhead of the function closure is likely higher than inlining (untested). In fact, this was why some of this kind of functionality got removed right before the 1.21 release. I think the Go community is going to prefer returning an iterator for these kinds of things, in which case you'd end up with a very similar number of LOC as if you do the straight evaluation inline for small data sets.

sethvargo:
> That's fair, but I do think these functions will be added to the standard library at some point. I'd rather not have to wait until they are added in order to use them. When/if they are added to the standard library it should be trivial to migrate to the standard library version as the function signatures will be very similar if not identical.

Also, FWIW, some of these _are_ in the standard library: https://pkg.go.dev/slices, https://pkg.go.dev/maps.

lock14:
> Hey @lock14 - thanks for doing this PR. The tests look great and the style of your code is correct. However, I'm not supportive of this change, largely for the reasons @stytchiz has pointed out.
> 
> This is not idiomatic Go and we should aim to align as closely with the community as possible. There are plans to add higher-order operations on collections into the standard library (we got a few in 1.21, but more are supposed to be coming), at which point we can consider something like this.
> 
> In general, these kinds of list comprehensions only make sense on small data structures where the cost of allocation or storage is low. In those situations, the overhead of the function closure is likely higher than inlining (untested). In fact, this was why some of this kind of functionality got removed right before the 1.21 release. I think the Go community is going to prefer returning an iterator for these kinds of things, in which case you'd end up with a very similar number of LOC as if you do the straight evaluation inline for small data sets.

Ok, that makes sense. I will close this PR for now then.

PR #231, opened 2023-10-27
feat: add scancode action and reusable workflow

sqin2019:
Goal: A Github action/workflow to run on workflow_dispatch on the entire repo to get all licenses, dependencies, and copyrights in a json and csv file.

### DIFF
--- a/.github/workflows/scancode.yml
+++ b/.github/workflows/scancode.yml
@@ -1,0 +1,54 @@
+# Copyright 2023 The Authors (see AUTHORS file)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# reusable workflow to call scancode action to scan licenses and copyrights and
+# upload the results.
+# TODO: add more usage guide here.
+name: 'scancode'
+on:
+  workflow_call:
+    inputs:
+      branch:
+        description: 'Which branch to scan.'
+        required: false
+        type: 'string'
+        default: 'main'
+env:
+  REPO_ROOT: '${{ github.repository }}'
+jobs:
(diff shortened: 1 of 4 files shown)


.github/workflows/scancode.yml
```diff
@@ -0,0 +1,53 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# reusable workflow to call scancode action to scan licenses and copyrights and
+# upload the results.
+# TODO: add more usage guide here.
+name: 'scancode'
```
sqin2019, line 18: We will need a caller workflow in the project repo, and the result will look like below, where you can download the json and csv file in your laptop:
Screenshot 2023-10-27 at 10 31 00 AM

.github/workflows/scancode.yml
```diff
@@ -0,0 +1,54 @@
...
+on:
+  workflow_call:
+    inputs:
+      branch:
```
sqin2019, line 22: This is for user to select a different branch to scan, for example I want to scan all the dependencies too, I can 'go mod vendor' (which download all vendor pkgs in the repo), then push it to a remote branch. On the GitHub UI, I can run the workflow to scancode the remote branch instead of main.

scancode/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM python:3.10
+
+ENV SCANCODE_RELEASE=32.0.8
+
+RUN apt-get update && apt-get install -y bzip2 xz-utils zlib1g libxml2-dev libxslt1-dev
```
sailorlqh, line 19: Just curious, not an opinion. Would it make sense if we build a base image with all these lib installed, and just build from that image. Would it save some time for building images? Is there any specific concerns for why we don't do that?

sqin2019: I tested that, the first one is the current way, and the second it use a base image, there is not much difference, plus we will have to maintain WIF and all that. Screenshot 2023-10-27 at 10 47 19 AM
Screenshot 2023-10-27 at 10 46 56 AM

sailorlqh (approved):

PR #241, opened 2023-11-06
Allow create tag on branches that are not default branch

### LINKED ISSUE #215: Remove create-tag branch condition on default branch
To prevent creating tags on unreviewed tag, we now only allow tags on repo default branch (main) since

1. we rely on deployment to enforce workflow review
2. but currently it does not prevent self-review. 

Once prevent self-review is done, see tracking issue https://github.com/github/roadmap/issues/825, we can remove the condition.

sqin2019:
fix #215 

prevent self-review on deployment has shipped according to https://github.com/github/roadmap/issues/825, it's now on deployment reviewer's hand to check if the tag created on unprotected branches is for valid reasons.

### DIFF
--- a/.github/workflows/create-tag.yml
+++ b/.github/workflows/create-tag.yml
@@ -169,12 +169,6 @@ jobs:
             const annotatedTag = process.env.ANNOTATED_TAG
             const message = process.env.MESSAGE
 
-            // TODO(#215): Once prevent self-review is supported by GitHub, we can remove this condition.
-            if (branch != '${{ github.event.repository.default_branch }}') {
-              core.setFailed(`branch (${branch}) is not the repository default branch (${{ github.event.repository.default_branch }}).`)
-              return
-            }
-
             let sha = '${{ github.sha }}'
             // If the input branch is not the branch of the calling workflow,
             // get the sha of input branch instead.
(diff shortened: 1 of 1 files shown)


lock14 (dismissed):

verbanicm (approved):

.github/workflows/create-tag.yml
```diff
@@ -169,12 +169,6 @@ jobs:
             const annotatedTag = process.env.ANNOTATED_TAG
             const message = process.env.MESSAGE
 
-            // TODO(#215): Once prevent self-review is supported by GitHub, we can remove this condition.
-            if (branch != '${{ github.event.repository.default_branch }}') {
-              core.setFailed(`branch (${branch}) is not the repository default branch (${{ github.event.repository.default_branch }}).`)
-              return
-            }
-
```
sethvargo, line 177: I would still like this to spit out a warning for reviewers.

sethvargo (changes requested):

verbanicm (approved):

.github/workflows/create-tag.yml
sqin2019: done, thanks!

sethvargo (approved):

PR #246, opened 2023-11-17
Only delete pipeline if it exists

lock14:
* Changed action to only call the delete command if the pipeline already exists.
* Made commmand line flag assignments consistent by always using '='
* Made script indentation consistent

### DIFF
--- a/.github/actions/deploy-dataflow-pipeline/action.yml
+++ b/.github/actions/deploy-dataflow-pipeline/action.yml
@@ -106,24 +106,24 @@ runs:
         GCS_TEMPLATE_FILE="gs://${BUCKET_NAME}/templates/${TEMPLATE_FILE_NAME}"
 
         gcloud dataflow flex-template build "${GCS_TEMPLATE_FILE}" \
-          --project "${PROJECT}" \
-          --image "${IMAGE}" \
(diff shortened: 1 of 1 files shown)


.github/actions/deploy-dataflow-pipeline/action.yml
```diff
@@ -106,24 +106,30 @@ runs:
...
-          --project "${PROJECT}" \
-          --image "${IMAGE}" \
-          --sdk-language "${PIPELINE_SDK}" \
-          --metadata-file "${TEMPLATE_METADATA_FILE}"
+          --project="${PROJECT}" \
+          --image="${IMAGE}" \
+          --sdk-language="${PIPELINE_SDK}" \
+          --metadata-file="${TEMPLATE_METADATA_FILE}"
 
-        gcloud beta datapipelines pipeline delete "${PIPELINE_NAME}" \
```
sethvargo, line 114: This is a race condition. It's possible for the pipeline to be created between this call and the next. I think we should always delete, but check the returned error message and ignore it if it's "not found".

lock14: Changed to ignore failure to delete.

sethvargo (approved):

PR #255, opened 2024-01-06
feat: make go test work for windows

capri-xiyue:
make go test work for windows.
Currently go test doesn't work for windows, see https://github.com/abcxyz/abc/actions/runs/7427886424/job/20214319302?pr=348

### DIFF
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -58,7 +58,7 @@ jobs:
       - name: 'Configure environment variables'
         if: '${{ inputs.env != ''{}'' }}'
         run: |-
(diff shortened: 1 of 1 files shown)


yolocs:
I suppose windows would require powershell instead.

.github/workflows/go-test.yml
```diff
@@ -58,7 +58,7 @@ jobs:
       - name: 'Configure environment variables'
         if: '${{ inputs.env != ''{}'' }}'
         run: |-
-          jq -r "to_entries | map(\"\(.key)=\(.value|tostring)\") | .[]" <<< '${{ inputs.env }}' >> $GITHUB_ENV
```
sethvargo, line 61: I think if you set `shell: 'bash'` this will just work as-is

sethvargo (approved):

PR #269, opened 2024-02-01
Add chmod +x (executable) to setup-binary action

kevya-google:
Sometimes the downloaded binary wasn't marked as executable (https://github.com/abcxyz/grc/pull/318/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR225).

This fixes the need to have an additional step to mark it with the proper permission.

### DIFF
--- a/.github/actions/setup-binary/action.yml
+++ b/.github/actions/setup-binary/action.yml
@@ -34,6 +34,9 @@ inputs:
     description: 'Add the install_path to the $PATH variable.'
     default: 'false'
     required: false
+  mark_executable:
+    description: 'Name of binary to mark as executable (chmod +x) in install_path. If this value is empty, no chmod will occur.'
+    required: false
 
 runs:
   using: 'composite'
@@ -132,9 +135,6 @@ runs:
 
         "${GOGETTER_PATH}/go-getter" "${DOWNLOAD_URL}" "${INSTALL_PATH}"
 
-        # Mark binary as executable
-        chmod +x ${INSTALL_PATH}/*
-
     - name: 'Save cache binary'
       if: |
         inputs.cache_key != '' && steps.cache-binary.outputs.cache-hit != 'true'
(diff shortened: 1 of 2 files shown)


sethvargo:
Is there a reason we wouldn't want to always do this?

.github/actions/setup-binary/action.yml
```diff
@@ -150,3 +153,13 @@ runs:
...
+
+    - name: 'Mark binary as executable'
+      if: |
+        inputs.mark_executable != ''
+      shell: 'bash'
+      env:
+        INSTALL_PATH: '${{ inputs.install_path }}'
+        MARK_EXECUTABLE: '${{ inputs.mark_executable }}'
+      run: |-
+        chmod +x ${INSTALL_PATH}/${MARK_EXECUTABLE}
```
verbanicm, line 165: The binary could be nested in the install path, the name mark_executable is kind of confusing. Maybe we separate this in to two inputs

`set_executable` - default false
`binary_subpath` - only used if set_executable = true

See https://github.com/abcxyz/pkg/blob/5c6ce400b0f1f7fc315e0cc5a8e9f6699fe6ef16/.github/actions/setup-binary/action.yml#L145 on testing for truthy values

kevya-google:
> Is there a reason we wouldn't want to always do this?

Not that I can think of, but maybe the flexibility is good? I can make this the default behavior.

.github/actions/setup-binary/action.yml
kevya-google: Separated into two inputs.

verbanicm:
I guess we could do it all the time, just wanted to allow user to choose.

kevya-google:
> I guess we could do it all the time, just wanted to allow user to choose.

In that case, should I make binary_subpath required?

.github/actions/setup-binary/action.yml
```diff
@@ -24,6 +24,9 @@ inputs:
   install_path:
     description: 'The path on disk in which to extract the binary, this cannot be equal to the current directory. Set `add_to_path` to true to add this to the $PATH.'
     required: true
+  binary_subpath:
+    description: 'Name of binary. Needed to set as executable.'
```
verbanicm, line 28: this should probably be more descriptive

```suggestion
    description: 'The subpath to the binary relative to the install path. This should include the binary name (e.g. dist/binary-name). Required to set the binary as executable.'
```

verbanicm (approved):

sethvargo (approved):

verbanicm:
Merging so I can use this....

PR #270, opened 2024-02-02
Add optional binary rename input to setup-binary action

kevya-google:
Merge after https://github.com/abcxyz/pkg/pull/269

### DIFF
--- a/.github/actions/setup-binary/action.yml
+++ b/.github/actions/setup-binary/action.yml
@@ -37,6 +37,9 @@ inputs:
     description: 'Add the install_path to the $PATH variable.'
     default: 'false'
     required: false
+  binary_rename:
+    description: 'New name of binary to rename to.'
+    required: false
 
 runs:
   using: 'composite'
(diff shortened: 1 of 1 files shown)


sethvargo:
Feels like we want to name this "destination"? It's not just the "name", but also the entire path.

.github/actions/setup-binary/action.yml
```diff
@@ -37,6 +37,9 @@ inputs:
     description: 'Add the install_path to the $PATH variable.'
     default: 'false'
     required: false
+  destination_subpath:
+    description: 'New destination of binary subpath (i.e. renaming).'
```
verbanicm, line 41:
```suggestion
    description: 'The destination subpath of the binary relative to the install path. This should include the binary name (e.g. dist/binary-name) and is useful for renaming the binary after download.'
```

.github/actions/setup-binary/action.yml
verbanicm, line 1: Should we validate that binary_subpath has a value in the "Validate inputs" step?

kevya-google:
> As a follow up, we should probably update the test workflow to use the destination_subpath

See: https://github.com/abcxyz/pkg/pull/271

PR #273, opened 2024-02-06
Update add_to_path to include binary_subpath

kevya-google:
This is needed in the case that the binary does not sit directly in the install_path.

### DIFF
--- a/.github/actions/setup-binary/action.yml
+++ b/.github/actions/setup-binary/action.yml
@@ -158,10 +158,11 @@ runs:
       shell: 'bash'
       env:
         INSTALL_PATH: '${{ inputs.install_path }}'
+        BINARY_SUBPATH: '${{ inputs.binary_subpath }}'
       run: |-
         # Add binary to path
-        echo "Adding ${INSTALL_PATH} to PATH"
-        echo "${INSTALL_PATH}" >> $GITHUB_PATH
+        echo "Adding ${INSTALL_PATH}/${BINARY_SUBPATH} to PATH"
+        echo "${INSTALL_PATH}/${BINARY_SUBPATH}" >> $GITHUB_PATH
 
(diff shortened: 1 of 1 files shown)


.github/actions/setup-binary/action.yml
```diff
@@ -158,10 +158,11 @@ runs:
       shell: 'bash'
       env:
         INSTALL_PATH: '${{ inputs.install_path }}'
+        BINARY_SUBPATH: '${{ inputs.binary_subpath }}'
       run: |-
         # Add binary to path
-        echo "Adding ${INSTALL_PATH} to PATH"
-        echo "${INSTALL_PATH}" >> $GITHUB_PATH
+        echo "Adding ${INSTALL_PATH}/${BINARY_SUBPATH} to PATH"
```
verbanicm, line 164: Do we need to consider destination_subpath at all? We might want to move this to the end consider which to add to path.

kevya-google: That's a good point, let me think how to re-organize the steps.

kevya-google: Ok, I think this should work:
1. MV install_path / binary_subpath -> install_path / destination_subpath
2. If destination_subpath is present, binary_subpath = destination_subpath.
    Add install_path / binary_subpath to PATH.
4. If destination_subpath is present, binary_subpath = destination_subpath.
    Mark install_path / binary_subpath as executable.

lock14 (commented):
What is the purpose of installing the binary to a different location than `INSTALL_PATH`?

kevya-google:
> What is the purpose of installing the binary to a different location than `INSTALL_PATH`?

destination_subpath exists https://github.com/abcxyz/pkg/pull/270 to allow for binary renaming.

verbanicm (approved):

Issue #259, opened 2024-01-16
Add github app auth functionality for repo-level requests


dcreey:
### TL;DR

[This](https://github.com/abcxyz/github-metrics-aggregator/blob/main/pkg/githubauth/app.go) should be migrated to this repo.

PR #276, opened 2024-02-10
Unify github auth and app management

sethvargo:
This merges GitHub auth and GitHub app management into a single package. I POCed this on GMA, and it provides a much nicer API and abstraction

Closes https://github.com/abcxyz/pkg/issues/259

/cc @dcreey


githubauth/app.go
```diff
@@ -0,0 +1,380 @@
...
+		return "", fmt.Errorf("invalid access token from GitHub - Body: %s", string(b))
+	}
+	return string(b), nil
+}
+
+// Token adheres to the oauth2 TokenSource interface and returns a oauth2 token
+// by creating a JWT token.
+func (g *App) OAuthTokenSource() oauth2.TokenSource {
+	return oauth2TokenSource(func() (*oauth2.Token, error) {
+		jwt, err := g.AppToken()
```
verbanicm, line 277: nit: can we name this more clearly? e.g. OAuthAppTokenSource. Its not immediately clear that you get the app token here and not the installation access token.

sethvargo: Done.

verbanicm (approved):

dcreey:
Nice! Thank you

PR #278, opened 2024-02-14
Remove deprecated libraries

sethvargo:
This will require a major version bump of pkg


pdewilde (approved):
Got rid of a lot of dependencies!

verbanicm (approved):

PR #280, opened 2024-02-14
make `LookupEnv` use cached `v` parameter

lock14:


sailorlqh (approved):

sethvargo (changes requested):
This is actually intentional; we don't lock, so we want to create the local allocation in case the underlying variables changes beneath us. You cannot guarantee that the `c.stdout` on L1 is the same `c.stdout` that you return.

pdewilde:
> This is actually intentional; we don't lock, so we want to create the local allocation in case the underlying variables changes beneath us. You cannot guarantee that the `c.stdout` on L1 is the same `c.stdout` that you return.

The only place I can find that is setting stdout is Pipe()

pdewilde:
Oops, that wasn't the "delete draft button"

pdewilde (approved):

sethvargo (approved):

PR #282, opened 2024-02-15
Update all deps

sethvargo:


sethvargo:
Looks like CodeQL isn't updated yet

sethvargo:
Dropping the Go update so that I can get the template updates in

verbanicm (approved):

PR #289, opened 2024-02-17
Add cleanup scripts

sethvargo:


.github/workflows/cleanup.yml
```diff
@@ -0,0 +1,96 @@
...
+
+            let workflowRuns = [];
+            for(const repo of repos) {
+              const runs = await github.paginate(github.rest.actions.listWorkflowRunsForRepo, {
+                owner: context.repo.owner,
+                repo: repo.name,
+                created: `< ${cutoffString}`,
+                exclude_pull_requests: true,
+                per_page: 100,
+              });
```
verbanicm, line 72: You can just request all completed workflow runs, rather then looping over them:


```suggestion
const runs = await github.paginate(github.rest.actions.listWorkflowRunsForRepo, {
  owner: context.repo.owner,
  repo: repo.name,
  status: 'completed',
  created: `< ${cutoffString}`,
  exclude_pull_requests: true,
  per_page: 100,
});
```

sethvargo: I don't think that will include cancelled, skipped, failed, stale, or timed out runs? Per https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28, it's not super clear what "completed" includes?

verbanicm (approved):

verbanicm (approved):

PR #290, opened 2024-02-17
Add new release workflows

sethvargo:
This will also require updates to minty's config.


.github/workflows/release.yml
```diff
@@ -0,0 +1,51 @@
...
+on:
+  push:
+    branches:
+      - 'main'
+      - 'release/**/*'
+
+jobs:
+  release:
+    runs-on: 'ubuntu-latest'
+    if: '${{ startsWith(github.event.head_commit.message, ''Release: v'') && github.event.head_commit.author.email == vars.TOKEN_MINTER_GITHUB_EMAIL }}'
```
verbanicm, line 26: I prefer the following syntax for readability:

```suggestion
    if: |-
      startsWith(github.event.head_commit.message, 'Release: v') &&
      github.event.head_commit.author.email == vars.TOKEN_MINTER_GITHUB_EMAIL
```

.github/actions/create-release/action.yml
```diff
@@ -0,0 +1,140 @@
...
+          // entrypoint.
+          const matches = context.payload.head_commit.message.match(/Release: v(?<version>[^\ ]+)/i);
+          if (!matches || !matches.groups) {
+            core.setFailed(`Commit message does not contain a version`)
+            return;
+          }
+          let releaseVersion = matches.groups.version;
+          while(releaseVersion.charAt(0).toLowerCase() === 'v') {
+            releaseVersion = releaseVersion.substr(1);
+          }
```
verbanicm, line 113: Should this be an `if`?

.github/actions/create-release/action.yml
```diff
@@ -0,0 +1,140 @@
...
+          // Ensure the commit is signed.
+          const commitResult = await github.rest.repos.getCommit({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            ref: context.payload.head_commit.id,
+          })
+          if (!commitResult.data.commit.verification.verified) {
+            core.setFailed(`Commit is not signed`)
+            return;
+          }
```
verbanicm, line 100: Do we need this? I feel like the ruleset should enforce this?

.github/actions/create-release/action.yml
sethvargo: No, this handles `vvvvv1.2` -> `1.2`

.github/actions/create-release/action.yml
sethvargo: Probably not? When this was a reusable workflow, this was ensuring that the email address wasn't forged.

sethvargo: Actually, we do need this. It's possible to sign a commit as a bot (which will pass rulesets), but fail validation. This isn't just checking that the commit is signed, but it's also checking that the commit is _verified_; rulesets don't do that as far as I can tell.

verbanicm (approved):

verbanicm (approved):

PR #303, opened 2024-02-19
feat: add draft prs to create-pull-request

verbanicm:
This change add the ability to create pull requests in draft status.

This change also modifies the existing logic to enable auto merge.

Since both draft and auto-merge are binary states, this change allows the user to toggle the status when creating/updating the pull request and the actions will safely modify.

I ran into several issues with using github-script and the API, namely we have to do a bunch of additional validation against existing states, where the gh cli already does this. I opted for the CLI to not re-invent the wheel and allow for somewhat "cleaner" code.


sethvargo (approved):
Yea, using the CLI is nicer

PR #306, opened 2024-02-21
Switch to nicer quoting for YAML conditions

sethvargo:
This is a verbanicm@ trick, but using a multi-line avoids the double-quoting.

### DIFF
--- a/.github/workflows/go-lint.yml
+++ b/.github/workflows/go-lint.yml
@@ -94,7 +94,8 @@ jobs:
 
       - name: 'Lint (download default configuration)'
         id: 'load-default-config'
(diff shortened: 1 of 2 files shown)


verbanicm (commented):
You don't need the ${{}} when doing this

sethvargo:
The syntax highlighting is nicer though:

screenshot-20240220-IQbYQXQ2@2x

sethvargo:
Also, it's required when the result is expected to be a boolean:

screenshot-20240220-C9z4RPoC@2x

verbanicm (approved):

PR #309, opened 2024-02-27
Update to new release process

sethvargo:


.github/workflows/release.yml
```diff
@@ -58,3 +59,75 @@ jobs:
...
+      - name: 'Cleanup failed release'
+        env:
+          GH_TOKEN: '${{ steps.mint-token.outputs.token }}'
+          RELEASE_VERSION: 'v${{ needs.create-release.outputs.version }}'
+          REPO: '${{ github.repository }}'
+        run: |-
+          gh release delete "${RELEASE_VERSION}" \
+            --repo "${REPO}" \
+            --cleanup-tag \
+            --yes || true
```
verbanicm, line 133: I dont think this works: https://github.com/abcxyz/jvs/actions/runs/8056366451/job/22005366392#step:3:12

sethvargo: Ah - it's because the tag doesn't exist for drafts. I removed --cleanup-tag and it should work.

verbanicm (approved):

PR #321, opened 2024-03-22
create go-test action

shankiyani:


.github/actions/go-test/action.yml
```diff
@@ -0,0 +1,94 @@
...
+runs:
+  using: 'composite'
+  steps:
+    - name: 'Configure environment variables'
+      if: |-
+        ${{ inputs.env != '{}' }}
+      # Use bash shell here to make it work with Windows host runner.
+      shell: 'bash'
+      run: |-
+        jq -r "to_entries | map(\"\(.key)=\(.value|tostring)\") | .[]" <<< '${{ inputs.env }}' >> $GITHUB_ENV
```
sethvargo, line 59: This actually won't be required anymore, because you can se the env on the step/job and it will trickle down.

.github/actions/go-test/action.yml
```diff
@@ -0,0 +1,78 @@
...
+      uses: 'actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491' # ratchet:actions/setup-go@v5
+      with:
+        go-version: '${{ inputs.go_version }}'
+        go-version-file: '${{ inputs.go_version_file }}'
+
+    - name: 'Test'
+      shell: 'bash'
+      working-directory: '${{ inputs.directory }}'
+      run: |-
+        ${{ inputs.go_proxy }} test \
```
shankiyani, line 58: falls back to `go test`

.github/actions/go-test/action.yml
```diff
@@ -0,0 +1,78 @@
...
+    type: 'string'
+    default: 'go.mod'
+  go_packages:
+    description: 'List of go packages to test.'
+    type: 'string'
+    default: './...'
+  go_proxy:
+    description: 'Proxy script to wrap go commands.'
+    type: 'string'
+    default: 'go'
```
sethvargo, line 32: Maybe `go_command` would be better, and then the description can say "this can be a script..."

.github/actions/go-test/action.yml
sethvargo, line 58: This will handle any shell escaping for you (spaces, special characters, etc):

```suggestion
      working-directory: '${{ inputs.directory }}'
      env:
        GO_COMMAND: '${{ inputs.go_command }}'
      run: |-
        "${GO_COMMAND}" test \
```

.github/actions/go-test/action.yml
```diff
@@ -0,0 +1,78 @@
...
+      shell: 'bash'
+      working-directory: '${{ inputs.directory }}'
+      run: |-
+        ${{ inputs.go_proxy }} test \
+          -count=1 \
+          -coverprofile=./coverage.out \
+          -race \
+          -shuffle=on \
+          -timeout='${{ inputs.go_test_timeout }}' \
+          ${{ inputs.go_packages }}
```
sethvargo, line 64: Make these envvars like I suggested above. It prevents injection (it was less of a concern on a reusable workflow where we were the only customer.

.github/actions/go-test/action.yml
```diff
@@ -0,0 +1,78 @@
...
+    description: 'Timeout for Go tests.'
+    type: 'string'
+    default: '10m'
+  directory:
+    description: 'Directory to the root of the Go module.'
+    type: 'string'
+    default: '.'
+
+runs:
+  using: 'composite'
```
sethvargo, line 43: I'm skeptical a composite action is going to work for this use case. First, it makes the UI output weird (see "Difference 6: Logging"). Second, they don't inherit secrets, so you need to explicitly pass any secrets as inputs, which can be a lot of enumeration.

.github/actions/go-test/action.yml
```diff
@@ -0,0 +1,78 @@
...
+  steps:
+    - name: 'Checkout'
+      uses: 'actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11' # ratchet:actions/checkout@v4
```
sethvargo, line 46: This will overwrite anything previously checked out, including auth.

.github/actions/go-test/action.yml
```diff
@@ -0,0 +1,78 @@
...
+          -race \
+          -shuffle=on \
+          -timeout='${{ inputs.go_test_timeout }}' \
+          ${{ inputs.go_packages }}
+
+    - name: 'Coverage'
+      shell: 'bash'
+      working-directory: '${{ inputs.directory }}'
+      run: |-
+        OUTPUT="$(${{ inputs.go_proxy }} tool cover -func=./coverage.out)"
```
sethvargo, line 70: Same note about `env`.

.github/actions/go-test/action.yml
```diff
@@ -0,0 +1,78 @@
...
+    type: 'string'
+    default: 'go'
+  go_test_timeout:
+    description: 'Timeout for Go tests.'
+    type: 'string'
+    default: '10m'
+  directory:
+    description: 'Directory to the root of the Go module.'
+    type: 'string'
+    default: '.'
```
sethvargo, line 40: While we're here, it seems like the community has standardized on "kebab" instead of "snake" case: https://dev.to/n3wt0n/composite-actions-vs-reusable-workflows-what-is-the-difference-github-actions-11kd

PR #326, opened 2024-04-16
fix: performance of want_lgtm_all

verbanicm:
- The GitHub `contains` expression is not case sensitive which allows this to work, saving a job run and an API call [details](https://docs.github.com/en/actions/learn-github-actions/expressions#contains)
- The `edited` status for the review comment only triggers when a review comment body is updated, so it is not needed.

WANT_LGTM=all

### DIFF
--- a/.github/workflows/want-lgtm-all.yml
+++ b/.github/workflows/want-lgtm-all.yml
@@ -28,7 +28,6 @@ on:
     types:
       - 'submitted'
       - 'dismissed'
-      - 'edited'
   workflow_call:
 
 concurrency:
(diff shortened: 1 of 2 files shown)


.github/workflows/want-lgtm-all.yml
```diff
@@ -42,34 +41,13 @@ permissions:
...
-              return "false";
-            }
-
-            return "true";
-
-      - id: 'validate-reviews'
         name: 'Validate Reviews'
         if: |-
-          ${{ steps.want-lgtm.outputs.result == 'true' }}
+          contains(github.event.pull_request.body, 'want_lgtm=all')
```
bradegler, line 50: I think you are dropping the `toLowerCase`, unless that contains function does that already for you.

verbanicm: That function is not case sensitive FTW!

https://docs.github.com/en/actions/learn-github-actions/expressions#contains

bradegler (approved):

sethvargo (approved):

PR #329, opened 2024-06-08
feat: github app get access token for an installation given org

sqin2019:
rest api doc: https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#get-an-organization-installation-for-the-authenticated-app

required by https://github.com/abcxyz/team-link/issues/24

### DIFF
--- a/githubauth/app.go
+++ b/githubauth/app.go
@@ -43,7 +43,7 @@ const (
 	// URL used to retrieve GitHub App installation in an GitHub organization. The
 	// pattern must contain a single '%s' which represents where in the url to
 	// insert the GitHub organization name or id (not case sensitive).
-	defaultGitHubAppInstallationURLPattern = "https://api.github.com/orgs/%s/installation" //nolint
+	defaultGitHubAppInstallationURLPattern = "https://api.github.com/orgs/%s/installation"
 )
 
 const cacheKey = "github-app-jwt"
--- a/githubauth/app.go
+++ b/githubauth/app.go
@@ -27,6 +27,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strconv"
 	"time"
 
 	"golang.org/x/oauth2"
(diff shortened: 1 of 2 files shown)


sethvargo (commented):
Hi @sqin2019 - can you link to the GitHub documentation for this? I'm not sure I like overloading the `githubAccessToken` function like this.

sqin2019:
> Hi @sqin2019 - can you link to the GitHub documentation for this? I'm not sure I like overloading the `githubAccessToken` function like this.

Do you mean this [link](https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#get-an-organization-installation-for-the-authenticated-app)? 

Let me clarify a little bit more: 

- Goal - To be able to get an access token for a GitHub app installation that is installed on an different org.
- Step added in the `githubAccessToken` - get the app installation ID for that org (instead of the one in the *App object) and then use that ID to generate the access token.

Here are what I tried:
1. get the installation ID outside of this package - some of the code (`AppToken`) overlaps with this package. 
2. make `githubAccessToken` generic and can get any response given different GitHub request - essentially the same as this PR.

Do you have any suggestions?

sethvargo:
> Goal - To be able to get an access token for a GitHub app installation that is installed on an different org.

The installation ID is a globally unique string. As long as you have the installation ID, everything should just work. Are you seeing something different?

sqin2019:
> > Goal - To be able to get an access token for a GitHub app installation that is installed on an different org.
> 
> The installation ID is a globally unique string. As long as you have the installation ID, everything should just work. Are you seeing something different?

ah, I thought we can install the same private github app in different organizations if under the same enterprise, but it seems that a private github app is restricted to it's own org? I wonder what teamsync did as I don't see them using a github app per org.

sethvargo:
> Update - seems can easily share the same github app across orgs by making it public in the github enterprise cloud account, or by https://docs.github.com/en/apps/sharing-github-apps/making-your-github-app-available-for-github-enterprise-server if it is a enterprise server account.

Ah yes - annoyingly GitHub Apps are either "private" (only available in your org) or "public" (visible to the entire Internet); there is no option to selectively expose. So yes, we'd have to make the GitHub App public to use it outside of our organization.

sqin2019:
> > Update - seems can easily share the same github app across orgs by making it public in the github enterprise cloud account, or by https://docs.github.com/en/apps/sharing-github-apps/making-your-github-app-available-for-github-enterprise-server if it is a enterprise server account.
> 
> Ah yes - annoyingly GitHub Apps are either "private" (only available in your org) or "public" (visible to the entire Internet); there is no option to selectively expose. So yes, we'd have to make the GitHub App public to use it outside of our organization.

This githubauth package is designed for a single github app installation, should we make it support other installations too?

sethvargo:
I'm not sure what you mean?

sqin2019:
see https://github.com/abcxyz/pkg/pull/330

PR #330, opened 2024-06-12
Expand githubauth to pull installation IDs given an org, repo, or user

### LINKED ISSUE #331: [githubauth] Remove jwt caching
### TL;DR

We don't make http calls for minting the tokens anymore (it's done with the private key), so there's no real value in the cache. 

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

sethvargo:
This is a breaking API change to account for the added functionality.

Fixes #331

/cc @sqin2019

### DIFF
--- a/githubauth/app.go
+++ b/githubauth/app.go
@@ -27,75 +27,45 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 	"time"
 
 	"golang.org/x/oauth2"
-
-	"github.com/abcxyz/pkg/cache"
 )
 
-// URL used to retrieve access tokens. The pattern must contain a single '%s'
-// which represents where in the url to insert the installation id.
-const defaultGitHubAccessTokenURLPattern = "https://api.github.com/app/installations/%s/access_tokens" //nolint
-
-const cacheKey = "github-app-jwt"
+// defaultBaseURL is the default API base URL.
+const defaultBaseURL = "https://api.github.com"
 
 // App is an object that can be used to generate application level JWTs or to
 // request an OIDC token on behalf of an installation.
 type App struct {
-	AppID          string
-	InstallationID string
-	PrivateKey     *rsa.PrivateKey
+	appID      string
+	privateKey *rsa.PrivateKey
 
-	accessTokenURLPattern string
-
-	jwtTokenExpiration time.Duration
-	jwtCacheDuration   time.Duration
-	jwtCache           *cache.Cache[[]byte]
-
+	baseURL    string
 	httpClient *http.Client
+}
+
+// AppID returns the GitHub App's ID.
+func (a *App) AppID() string { //nolint:stylecheck // "AppID" is the name GitHub uses
+	return a.appID
 }
 
 // Option is a function that provides an option to the GitHub App creation.
 type Option func(g *App) *App
 
-// WithAccessTokenURLPattern allows overriding of the GitHub api url that is
-// used when generating installation access tokens. The default is the primary
-// GitHub api url which should only be overridden for private GitHub
-// installations.
-//
-// The `pattern` parameter expects a single `%s` that represents the
-// installation id that is provided with the rest of the configuration.
-func WithAccessTokenURLPattern(pattern string) Option {
+// WithBaseURL allows overriding of the GitHub API url. This is usually only
+// overidden for testing or private GitHub installations.
+func WithBaseURL(url string) Option {
 	return func(g *App) *App {
-		g.accessTokenURLPattern = pattern
-		return g
-	}
-}
-
-// WithJWTTokenExpiration is an option that allows overriding the default
-// expiration date of the application JWTs.
-func WithJWTTokenExpiration(exp time.Duration) Option {
-	return func(g *App) *App {
-		g.jwtTokenExpiration = exp
-		return g
-	}
-}
-
-// WithJWTTokenCaching is an option that tells the GitHub app to cache its JWT
-// App tokens. The amount of time that the tokens are cached is based on the
-// provided `beforeExp` parameter + the configured token expiration. This
-// results in a cache expiration of <token expiration> - <beforeExp>.
-func WithJWTTokenCaching(beforeExp time.Duration) Option {
-	return func(g *App) *App {
-		exp := g.jwtTokenExpiration
-		g.jwtCacheDuration = exp - beforeExp
+		g.baseURL = strings.TrimSuffix(url, "/")
 		return g
 	}
 }
 
 // WithHTTPClient is an option that allows a consumer to provider their own http
-// client implementation.
+// client implementation. This HTTP client will be shared among all
+// [AppInstallation].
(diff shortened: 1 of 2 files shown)


githubauth/app.go
```diff
@@ -103,17 +98,85 @@ func WithHTTPClient(client *http.Client) Option {
...
+		return u, nil
+	}
+}
+
 // NewApp creates a new GitHub App from the given inputs.
 //
 // The privateKey can be the [*rsa.PrivateKey], or a PEM-encoded string (or
 // []byte) of the private key material.
-func NewApp[T *rsa.PrivateKey | string | []byte](appID, installationID string, privateKeyT T, opts ...Option) (*App, error) {
+func NewApp[T *rsa.PrivateKey | string | []byte](ctx context.Context, appID string, privateKeyT T, accessTokensURLFunc AccessTokensURLFunc, opts ...Option) (*App, error) {
```
sqin2019, line 175: Can we make `AccessTokensURLFunc` optional and provide a func that can return an access token given `AccessTokensURLFunc`, do you see any concerns with that?

sethvargo: I'm not sure what you mean - can you show me an example?

sqin2019:
```
// Create an App without access token URL
func NewApp[T *rsa.PrivateKey | string | []byte](ctx context.Context, appID string, privateKeyT T, opts ...Option) (*App, error) {...}

// Use the given accessTokensURLFunc to get access token instead of g.accessTokensURL
func (g *App) AccessTokenWithURL(ctx context.Context, request *TokenRequestAllRepos/TokenRequest, accessTokensURLFunc AccessTokensURLFunc) (string, error) {...}
```

sqin2019: So that we can use the same `g *App` instance to get access tokens for any installations.

githubauth/app.go
```diff
@@ -103,17 +98,85 @@ func WithHTTPClient(client *http.Client) Option {
...
+		return "", fmt.Errorf("failed to create http request: %w", err)
+	}
+	req.Header.Set("Authorization", "Bearer "+jwt)
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("failed to make http request: %w", err)
+	}
+	defer resp.Body.Close()
+	if got, want := resp.StatusCode, 200; got != want {
```
yolocs, line 117: nit, I'd choose:
```suggestion
	if got, want := resp.StatusCode, http.StatusOK; got != want {
```

yolocs (approved):

githubauth/app.go
sethvargo: Hmm I see. The problem is that the downstream things (OauthAppTokenSource, githubAccessToken) don't accept those inputs (and they call AppToken). 

We could do something like 

```go
app := NewApp(...)

app.ForInstallation(...)
```

But the cost of creating apps is pretty low, and I'm not sure where we'd need multiple in the same scope?

sqin2019: It will be more convenient if I can use a single app instance to get access token given org in order to handle events for different orgs, avoid to create multiple app instances and they use the same cache and http client settings. 

For downstream funcs, it shouldn't be a problem to add the additional accessTokensURLFunc input, App JWT (AppToken) is the same if the github app is the same (IIUC).

sqin2019: To add, it also makes more sense for a service to manage just one github app instead of one per installation.

sethvargo: Hmm - the problem with that approach is that you're resolving the org -> installation_id on _each_ API call. I think we really want to separate "get an installation id" from "get an access token for that installation id". 

I'm also looking at [how go-github does this](https://github.com/google/go-github/blob/master/github/apps.go), and they separate the "app" from the installation, but you still create a separate *Installation for each connection.

So I think the pattern of App -> (many) Installation is correct. We can either:

1. Update to match go-github (you'd create 1 app, then have to mint N installations)
2. Use this code which (to be honest) I don't like because it complicates the code to support a single use case
3. Use go-github instead TeamSync

sqin2019: Thanks, very helpful, let me try 3 and user go-github directly to get access tokens.

sethvargo: @sqin2019 - I talked to @verbanicm about this yesterday, and he mentioned there are other use cases for supporting multiple orgs, so let me re-think this design. I think we need to separate "app" from "installation". I've got a plane ride today where I can carve out some time to POC this.

sqin2019: Sounds good, just curious, after trying it out with `go-github`, it seems the only things not covered in `go-github` are parsing the github app private key and mint the app JWT with it, is the purpose of `githubauth` in our pkg repo a light weight version of `go-github`?

sethvargo: Is the purpose of githubauth in our pkg repo a light weight version of go-github?

Yes, with less deps.

sethvargo: @sqin2019 PTAL

yolocs (approved):

githubauth/app.go
sethvargo: @sqin2019 I enabled auto-merge, so if you're okay with this approach, resolve this comment and the PR will merge. If you still have concerns, please keep this thread open.

githubauth/app.go
```diff
@@ -177,45 +135,128 @@ type TokenRequestAllRepos struct {
...
 	}
 
-	return token, nil
+	return unsigned + "." + b64Encode(signature), nil
 }
 
 // OAuthAppTokenSource adheres to the oauth2 TokenSource interface and returns a oauth2 token
 // by creating a JWT token.
-func (g *App) OAuthAppTokenSource() oauth2.TokenSource {
+func (i *App) OAuthAppTokenSource() oauth2.TokenSource {
```
sqin2019, line 174: nit: `a` instead of `i`?

githubauth/app.go
```diff
@@ -131,15 +99,10 @@ func NewApp[T *rsa.PrivateKey | string | []byte](appID, installationID string, p
 	}
 
 	app := &App{
-		AppID:          appID,
-		InstallationID: installationID,
-		PrivateKey:     privateKey,
-
-		accessTokenURLPattern: defaultGitHubAccessTokenURLPattern,
-
-		jwtTokenExpiration: 9 * time.Minute,
```
sqin2019, line 140: nit: should we keep this?

sqin2019 (approved):
Nice, thank you!

githubauth/app.go
sqin2019: Just a small one for the jwt expiration.

githubauth/app.go
sqin2019: nvm, I am fine with removing on a second thought.

githubauth/app.go
sethvargo: https://github.com/abcxyz/pkg/pull/332

PR #337, opened 2024-06-26
checkout config from default branch in load-workflow-variables action

gjonathanhong:
This change ensures we check out an approved config file prior to loading the workflow variables.

### DIFF
--- a/.github/actions/load-workflow-variables/action.yml
+++ b/.github/actions/load-workflow-variables/action.yml
@@ -37,7 +37,7 @@ runs:
       uses: 'actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11' # ratchet:actions/checkout@v4
       with:
         ref: '${{ github.event.repository.default_branch }}'
-        path: '${{ github.sha }}'
+        path: 'load-workflow-variables/${{ github.sha }}'
         sparse-checkout: |
           ${{ inputs.filepath }}
         sparse-checkout-cone-mode: 'false'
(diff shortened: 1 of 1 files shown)


kevya-google (approved):

.github/actions/load-workflow-variables/action.yml
```diff
@@ -33,11 +33,20 @@ inputs:
 runs:
   using: 'composite'
   steps:
+    - name: 'Get Approved Configuration File'
+      uses: 'actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11' # ratchet:actions/checkout@v4
+      with:
+        ref: '${{ github.event.repository.default_branch }}'
+        path: '${{ github.sha }}'
```
verbanicm, line 40: Should we prefix this with the action name to prevent collisions? Or use a tmp dir?

.github/actions/load-workflow-variables/action.yml
```diff
@@ -33,11 +33,20 @@ inputs:
...
+        sparse-checkout: |
+          ${{ inputs.filepath }}
+        sparse-checkout-cone-mode: 'false'
+
     - name: 'Load Workflow Variables'
       shell: 'bash'
       env:
         WORKING_DIRECTORY: '${{ inputs.working_directory }}'
-        FILEPATH: '${{ inputs.filepath }}'
+        FILEPATH: '${{ github.sha }}/${{ inputs.filepath }}'
```
verbanicm, line 49: I thnk path prefix needs to exist here too

.github/actions/load-workflow-variables/action.yml
gjonathanhong: opted for the former. The Checkout action restricts the destination path to within the current working directory. Also cannot change the `working_directory` when using `workflow_call` 👎

PR #339, opened 2024-07-02
Fix IntersectMapKeys first-wins bug

drevell:
There was a bug in the old version. The intended behavior was for the the map value to come from the first map in the list. However, the actual behavior was that the map value came from either the first map or from the smallest map, depending on the size order of the input maps.

### DIFF
--- a/sets/maps.go
+++ b/sets/maps.go
@@ -14,6 +14,10 @@
 
 package sets
 
+import (
+	gomaps "maps"
+)
+
 // IntersectMapKeys finds the intersection of all m keys, where intersection is
 // defined as keys that exist in all m. In the case where duplicate keys exist
 // across maps, the value corresponding to the key in the first map is used. It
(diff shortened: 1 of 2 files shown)


drevell:
Converted to draft while I rethink this algorithm

pdewilde:
> Converted to draft while I rethink this algorithm

You could still have the smallest map optimization.

```psuedocode
candidates = clone(smallestMap)
output = map[][]
for key in candidates:
  value = maps[0][key]
  for map in maps[1:] && value != null :
     if key not in map:
       value = null
   if value != null 
     output[key] = value
     
```

sethvargo:
What's the bug here?

pdewilde:
> What's the bug here?

From first revision of dave's description:

There was a bug in the old version. The intended behavior was for the the map value to come from the first map in the list. However, the actual behavior was that the map value came from either the first map or from the smallest map, depending on the size order of the input maps.

drevell:
> > What's the bug here?
> 
> From first revision of dave's description:
> 
> There was a bug in the old version. The intended behavior was for the the map value to come from the first map in the list. However, the actual behavior was that the map value came from either the first map or from the smallest map, depending on the size order of the input maps.

Right. You can cherry-pick just the test changes to reproduce the bug.

sets/maps.go
```diff
@@ -15,63 +15,39 @@
...
-		// For each key in our smallest set, check if the key exists in the current
-		// map. If it does not, remove it from the map since it's not part of the
-		// intersection.
-		for k := range final {
-			if _, ok := m[k]; !ok {
-				delete(final, k)
-			}
-		}
+		// Since the output keys must exist in every input map, and the earliest
+		// one takes precedence, then the earliest map is always the zeroth map.
```
pdewilde, line 46: This comment makes less sense now that the function comment no longer describes it as earliest.

pdewilde (approved):

sets/maps.go
drevell: I think it still gives some good intuition though. I'd like to leave it.

sethvargo:
https://github.com/abcxyz/pkg/pull/340

drevell:
> #340
> 
> I'm pretty sure the fix can be much simpler at the cost of one more iteration. Given the number of maps we're intersecting is usually the lower bound, this seems negligible.



> #340
> 
> I'm pretty sure the fix can be much simpler at the cost of one more iteration. Given the number of maps we're intersecting is usually the lower bound, this seems negligible.



> #340
> 
> I'm pretty sure the fix can be much simpler at the cost of one more iteration. Given the number of maps we're intersecting is usually the lower bound, this seems negligible.

From my point of view, the *fix* may be simpler in #340 (add a for loop), but the resulting algorithm is simpler in this PR.

sethvargo (approved):

PR #340, opened 2024-07-02
Fix a bug where IntersectMapKeys did not return the first value

sethvargo:
Closes https://github.com/abcxyz/pkg/pull/339

### DIFF
--- a/sets/maps.go
+++ b/sets/maps.go
@@ -44,8 +44,14 @@ func IntersectMapKeys[K comparable, V any](maps ...map[K]V) map[K]V {
 	}
 	smallestMap := maps[smallestIdx]
 	final := make(map[K]V, len(smallestMap))
-	for k, v := range smallestMap {
-		final[k] = v
+	for k := range smallestMap {
+		for _, m := range maps {
+			if _, ok := final[k]; !ok {
+				if fv, ok := m[k]; ok {
+					final[k] = fv
+				}
+			}
+		}
 	}
 
 	// Compute the intersection.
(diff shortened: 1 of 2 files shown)


sets/maps.go
```diff
@@ -44,8 +44,14 @@ func IntersectMapKeys[K comparable, V any](maps ...map[K]V) map[K]V {
 	}
 	smallestMap := maps[smallestIdx]
 	final := make(map[K]V, len(smallestMap))
-	for k, v := range smallestMap {
-		final[k] = v
+	for k := range smallestMap {
```
drevell, line 47: My impression is that this is perhaps over-elaborate compared to https://github.com/abcxyz/pkg/pull/339/files (5 for loops vs. 3). Is there a reason to prefer this approach?

pdewilde: Yeah, this is somewhat difficult to parse IMO.

sets/maps.go
```diff
@@ -44,8 +44,14 @@ func IntersectMapKeys[K comparable, V any](maps ...map[K]V) map[K]V {
...
+		for _, m := range maps {
```
pdewilde, line 48: There is no need to iterate over all maps, if a value is in the intersection, it will appear in the 1st map

sets/maps.go
```diff
@@ -44,8 +44,14 @@ func IntersectMapKeys[K comparable, V any](maps ...map[K]V) map[K]V {
 	}
 	smallestMap := maps[smallestIdx]
 	final := make(map[K]V, len(smallestMap))
```
pdewilde, line 46: I don't know if preallocating final with `len(smallestMap)`  is the right tradeoff or not. You are saving at most log(n) allocs and n copies, but possibly wasting n space in the scenario that the smallest map is large but the intersection is small. (Dave pointed this out in a discussion)

PR #348, opened 2024-09-27
Fix load-workflow-variables issue where config file cannot be found

gjonathanhong:
This change removes the `./` prefix from the input filepath before passing the value onto the following steps. If `./guardian.yml` is provided, then this will be sanitized to `guardian.yml`.

This fixes an issue with `load-workflow-variables` action caused by providing a relative path for `inputs.filepath`, e.g. `./guardian.yml` instead of `guardian.yml`. Passing this value to the `sparse-checkout` option of the Checkout step caused the action to not match with the `guardian.yml` file at the root of the repository, resulting in a file not found error in the `Load Workflow Variables` step.

### DIFF
--- a/.github/actions/load-workflow-variables/action.yml
+++ b/.github/actions/load-workflow-variables/action.yml
@@ -33,6 +33,9 @@ inputs:
 runs:
   using: 'composite'
   steps:
+    # Removes "./" from the input filepath, if exists. This avoids an issue with
+    # the sparse-checkout option in the Checkout step, where a relative path,
+    # e.g. "./env.yml", will not match with the expected "env.yml" file.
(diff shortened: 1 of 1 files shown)


.github/actions/load-workflow-variables/action.yml
```diff
@@ -33,20 +33,30 @@ inputs:
 runs:
   using: 'composite'
   steps:
+    - name: 'Sanitize Input Filepath'
```
sethvargo, line 36: It'd be helpful to document this step inline. There's good documentation in the PR, but we should assume that all that history is lost post-merge :)

sethvargo (approved):

PR #354, opened 2024-10-24
githubauth - support request signing by Cloud KMS

Celdir:
Currently GitHub requests are signed by a private key we store in an environment variable and pass into `githubauth.App` to sign the AppToken. This PR adds support for signing the AppToken with a private key stored in Cloud KMS instead, so we never have to handle the private key directly.


sethvargo (commented):
1. Are we actually able to import GitHub's private key into KMS? Cloud KMS only permit a submit of key format and algorithms.

1. I'm not convinced we want to make this API call to Cloud KMS in the hot path?

Celdir:
> 1. Are we actually able to import GitHub's private key into KMS? Cloud KMS only permit a submit of key format and algorithms.
> 2. I'm not convinced we want to make this API call to Cloud KMS in the hot path?

1. Cloud KMS does support `2048 bit RSA - PKCS#1 v1.5 padding - SHA256 Digest` which I believe is the format of the GitHub private key. 
2. Talked with Brad and Mike about this. They said you have concerns about a possible DDoS attack that causes us to hit KMS quota for minty on public repos, but Mike suggested only using KMS signing for alphabet enterprise orgs to mitigate that risk. We could use the `enterprise_id` from the GitHub token to secure the minty endpoint for only alphabet enterprise orgs. The relevant changes to the WIF pool and app configuration would be done in the minty repo. Does this sound reasonable?

PR #358, opened 2024-12-20
Multi in-org approvals for out-of-org PRs

ian-shafer:
GitHub reusable workflow to enforce code review requirements, namely that each PR should be reviewed by a member of the organization. This workflow prevents the use of "sock puppet" accounts to work around this requirement by ensuring that PRs originating from outside the org (from a potential sock puppet account) must be approved by at least two in-org members.

Note that we require a JSON file with all members because the token automatically provided to workflows does not allow access to orgs. See [Permissions for the GITHUB_TOKEN](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token).

This members JSON file is problematic because, by definition, the file is stale as soon as it's generated. There are two cases to consider:

1. The JSON file is missing a current member - Not that concerning. The missing member will not be counted as in in-org approver. At worst this will require more reviews.
1. The JSON file contains a member that was removed - Concerning (but no worse than today). The removed member will be counted as an in-org approver when they should not be counted as such.

The above concern can be alleviated by getting access to the orgs API to get real-time org membership information.

### DIFF
--- a/.github/workflows/multi-approvers.js
+++ b/.github/workflows/multi-approvers.js
@@ -1,0 +1,110 @@
+const APPROVED = 'APPROVED';
+const COMMENTED = 'COMMENTED';
+const MIN_APPROVED_COUNT = 2;
+
+/** Returns true if the login exists in the members list. */
+function containsLogin(members, login) {
+  return !!members.find((member) => member.login === login);
+}
+
+/** Returns the number of approvals from members in the given list. */
+function inOrgApprovedCount(members, submittedReviews, prLogin) {
+  const reviewStateByLogin = {};
+  submittedReviews
+    // Remove the PR user.
+    .filter((r) => r.user.login !== prLogin)
+    // Only consider users in the org.
+    .filter((r) => containsLogin(members, r.user.login))
+    // Sort chronologically ascending. Note that a reviewer can submit multiple reviews.
+    .sort((a, b) => new Date(a.submitted_at) - new Date(b.submitted_at))
+    .forEach((r) => {
+      const reviewerLogin = r.user.login;
+
+      // Set state if it does not exist.
+      if (!Object.hasOwn(reviewStateByLogin, reviewerLogin)) {
+        reviewStateByLogin[reviewerLogin] = r.state;
+        return;
+      }
+
+      // Always update state if not approved.
+      if (reviewStateByLogin[reviewerLogin] !== APPROVED) {
+        reviewStateByLogin[reviewerLogin] = r.state;
+        return;
+      }
+
+      // Do not update approved state for a comment.
+      if (reviewStateByLogin[reviewerLogin] === APPROVED && r.state !== COMMENTED) {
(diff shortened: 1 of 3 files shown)


.github/workflows/multi-approvers.js
```diff
@@ -0,0 +1,110 @@
...
+        reviewStateByLogin[reviewerLogin] = r.state;
+      }
+    })
+
+  return Object.values(reviewStateByLogin).filter((s) => s === APPROVED).length;
+}
+
+/** Checks that approval requirements are satisfied. */
+async function onPullRequest({orgMembersPath, prNumber, repoName, repoOwner, github, core}) {
+  const members = require(orgMembersPath);
```
pdewilde, line 46: Might make sense to marshal into a map. Will simplify and make containsLogin fast

ian-shafer: Done.

sethvargo:
Silly question, but what's the reason for specifying members as an input instead of querying the GitHub API directly?

ian-shafer:
> Silly question, but what's the reason for specifying members as an input instead of querying the GitHub API directly?

Not silly at all. The issue is that the token automatically provided to workflows does not allow access to orgs. See [Permissions for the GITHUB_TOKEN]([Permissions for the GITHUB_TOKEN](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token).

This PR is meant to be a v0 for this feature. In the future, we can get access to the org APIs using minty, installing a GitHub application, or  some other way that I don't know about. 

Also, note that using a JSON file with a members list that is stale (it's stale as soon as it's created, by definition) has two cases to consider:
1) The JSON file is missing a current member - Not that concerning. The missing member will not be counted as in in-org approver. At worst this will require more reviews.
2) The JSON file contains a member that was removed - Concerning (but no worse than today). The removed member will be counted as an in-org approver when they should not be counted as such.

I'll update the PR comment with this (I meant to initially but forgot).

pdewilde:
> 2. The JSON file contains a member that was removed - Concerning (but no worse than today). The removed member will be counted as an in-org approver when they should not be counted as such.
> 

Clearly we need to implement zanzibar to prevent this /s

sethvargo:
Thanks @ian-shafer. I'm worried about the case where someone submits a PR while modifying the `members.json` file to include themselves.

> Clearly we need to implement zanzibar to prevent this /s

This is "jzonibar"; totally different @pdewilde 😄

mattrandallbecker (approved):
Approval for Typescript readability. 

Re: sneaking into members, I think this is a good V0, even if someone proposed adding themselves to members it would require two members to approve it.

Once this version is out we can start to think about a token-minter version that would be more difficult to set up, but be GitHub team-based rather than config based.

sethvargo:
@mattrandallbecker agreed - I don't want to stand in the way of incremental progress.

PR #377, opened 2025-01-16
feat: support for using KMS for signing requests for GitHub apps

bradegler:


githubauth/app.go
```diff
@@ -89,21 +88,39 @@ func NewApp[T *rsa.PrivateKey | string | []byte](appID string, privateKeyT T, op
...
+		g.signer = NewPrivateKeySigner(privateKey)
+		return g, nil
+	}
+}
+
+// WithKMSSigner is one of two options that configures how requests
+// to GitHub are signed. This option utilizes Google Cloud KMS
+// via the provided keyID to sign the request. The keyID is in the
+// format "projects/*/locations/*/keyRings/*/cryptoKeys/*.".
+func WithKMSSigner(ctx context.Context, keyID string) Option {
```
sethvargo, line 109: Might want to say "GoogleCloudKMS" in case we need to support AWS KMS or Vault KMS in the future.

githubauth/app.go
```diff
@@ -113,7 +130,14 @@ func NewApp[T *rsa.PrivateKey | string | []byte](appID string, privateKeyT T, op
...
 
 	for _, opt := range opts {
-		app = opt(app)
+		app, err = opt(app)
+		if err != nil {
+			return nil, fmt.Errorf("error when applying option: %w", err)
+		}
+	}
+
+	if app.signer == nil {
```
sethvargo, line 139: It feels like we really want to have `NewApp` take a [`crypto.Signer`](https://pkg.go.dev/crypto#Signer) as a required argument. Then you could instantiate the signer outside of the call (by parsing an RSA key, setting up GCP KMS, etc).

bradegler: That's a reasonable suggestion, I'll try pivoting to that and see what it looks like.

bradegler: I think you are correct and this looks better.

sethvargo: It happens sometimes

capri-xiyue (approved):

Celdir (approved):

go.mod
```diff
@@ -5,12 +5,14 @@ go 1.22.0
 toolchain go1.23.4
 
 require (
+	cloud.google.com/go/kms v1.20.2
```
sethvargo, line 8: @bradegler Ooof - I'm worried about all these new dependencies. One of the advantages of using a `crypto.Signer` interface is that we can push this out into another package. This feels like a lot of dependencies to take/maintain across all of `pkg` customers just to avoid these LOC:

```go
	client, err := kms.NewKeyManagementClient(ctx)
	if err != nil {
		return nil, fmt.Errorf("failed to setup client: %w", err)
	}
	signer, err := gcpkms.NewSigner(ctx, client, keyID)
	if err != nil {
		return nil, fmt.Errorf("failed to create signer: %w", err)
	}
	return signer, nil
```

bradegler: That's fair. I just didn't feel like reimplementing this every time we used it but the extra dependencies do suck. I see you already fixed it.

sethvargo: Yea - I never complain without accompanying code :). Also, I just updated all the deps on go-gcpkms too.

Issue #364, opened 2025-01-15
Multi-approvers workflow to use list of public organization members


meltsufin:
### TL;DR

Currently, `multi-approvers` requires a `members.json` to list out all users who qualify for approving PRs. This is toilsome and error-prone to configure for each repository in the organization. There should be an option for the workflow to instead use the public list of organization members.

### Detailed design

```markdown

```

### Alternatives considered

```markdown

```

### Additional information

See: #361 for more context.

ian-shafer:
We are currently addressing this. I would expect a fix to be submitted within the next two weeks.

ian-shafer:
This issue is now resolved with https://github.com/abcxyz/actions/pull/35. This new GH Action does not support a members JSON file -- it uses GH teams to test membership.

PR #382, opened 2025-01-30
Support single members.json for entire org

meltsufin:
Allowing `members.json` to be downloaded from any repository, enables configuring multi-approvers as an org-level workflow with a single member list in one place.

### DIFF
--- a/.github/workflows/multi-approvers.yml
+++ b/.github/workflows/multi-approvers.yml
@@ -18,7 +18,7 @@ on:
   workflow_call:
     inputs:
       # Path to a JSON file containing the members of the org. This should be
-      # the full path from the repo root without a leading `/`.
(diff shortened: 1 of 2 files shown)


sethvargo:
I really think we need to focus on getting the org member list from the GitHub API instead of relying on a static JSON file.

meltsufin:
> I really think we need to focus on getting the org member list from the GitHub API instead of relying on a static JSON file.

Agreed. I was exploring this as a short-term solution until #364 is implemented.

mattrandallbecker (approved):

mattrandallbecker:
> I really think we need to focus on getting the org member list from the GitHub API instead of relying on a static JSON file.

Agree, and the long term plan is to still do that, but rolling out [github-token-minter](https://github.com/abcxyz/github-token-minter) broadly will take a lot more time.

PR #383, opened 2025-01-31
Add auto-compute paths to create-pull-request

sethvargo:
This removes the need to manually specify the filepaths that were changed or deleted. It's backwards compatible and requires opt-in via setting "compute_paths: true".

I added a fair amount of debug logging because I suspect there will be some trial and error here...

### DIFF
--- a/.github/actions/create-pull-request/action.yml
+++ b/.github/actions/create-pull-request/action.yml
@@ -47,6 +47,10 @@ inputs:
     description: 'Create pull request in draft status.'
     required: false
     default: 'false'
+  compute_paths:
+    description: 'Automatically compute the changed and deleted files. Defaults to `false`.'
+    required: false
+    default: 'false'
   changed_paths:
     description: 'JSON array of the relative file paths added or changed in the pull request. Defaults to `[]`.'
     required: false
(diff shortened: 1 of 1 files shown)


bradegler (approved):

.github/actions/create-pull-request/action.yml
```diff
@@ -189,17 +193,66 @@ runs:
...
+        if [ "${COMPUTE_PATHS}" = true ]; then
+          # Make a scratch directory that gets cleaned up.
+          SCRATCH="$(mktemp -d -p "${RUNNER_TEMP}" -t "create-pull-request-${GITHUB_SHA:0:7}")"
+          cleanup() { rm -rf "${SCRATCH}" }
+          trap cleanup EXIT
+
+          # Copy everything into the scratch directory so we don't mess with the
+          # local tree in case there are other actions that assume a pristine
+          # working directory.
+          cp -r "${PWD}" "${SCRATCH}"
```
pdewilde, line 214: Shouldn't this be  `cp -r "$(PWD)" "${SCRATCH}"` ?

I don't see anythign that is setting PWD as a variable

pdewilde: Nevermind, TIL PWD is also a variable given for free by bash

pdewilde, line 214: When I try this locally, I'm getting an extra layer of nesting:

```bash
pdewilde@pcd:/tmp/foo$ tree -a
.
├── .hidden
├── nested
│   ├── .nested_hidden
│   └── nested_normal
└── normal

2 directories, 4 files
pdewilde@pcd:/tmp/foo$ cp -r "${PWD}" "${SCRATCH}"
pdewilde@pcd:/tmp/foo$ cd "${SCRATCH}"
pdewilde@pcd:/tmp/create-pull-vCAv$ tree -a
.
└── foo
    ├── .hidden
    ├── nested
    │   ├── .nested_hidden
    │   └── nested_normal
    └── normal

3 directories, 4 files
```

.github/actions/create-pull-request/action.yml
```diff
@@ -189,17 +193,66 @@ runs:
...
+      id: 'compute-changes'
+      shell: 'bash'
+      env:
+        COMPUTE_PATHS: '${{ inputs.compute_paths }}'
+        CHANGED_PATHS: '${{ inputs.changed_paths }}'
+        DELETED_PATHS: '${{ inputs.deleted_paths }}'
+      run: |-
+        if [ "${COMPUTE_PATHS}" = true ]; then
+          # Make a scratch directory that gets cleaned up.
+          SCRATCH="$(mktemp -d -p "${RUNNER_TEMP}" -t "create-pull-request-${GITHUB_SHA:0:7}")"
```
pdewilde, line 207: I'm getting an error when trying this:

```
pdewilde@pcd:/tmp/foo$ GITHUB_SHA=ffac537e6cbbf934b08745a378932722df287a53
pdewilde@pcd:/tmp/foo$ SCRATCH="$(mktemp -d -p "${RUNNER_TEMP}" -t "create-pull-request-${GITHUB_SHA:0:7}")"
mktemp: too few X's in template ‘create-pull-request-ffac537’
```

I think the issue is this works on BSD mktemp but not the GNU version. See: https://unix.stackexchange.com/questions/30091/fix-or-alternative-for-mktemp-in-os-x

.github/actions/create-pull-request/action.yml
pdewilde: (found this because I was suspicious that dotfiles wouldn't be copied over, but seems like that isn't an issue)

.github/actions/create-pull-request/action.yml
pdewilde: Appears was fixed in https://github.com/abcxyz/pkg/pull/385

Issue #361, opened 2025-01-10
multi-approvers.yml requires revealing concealed organization members


meltsufin:
The `members.json` configuration for `multi-approvers.yml` workflow requires all organization members to be made public.
This is not desirable if there are good reasons for certain members to be private. On the other hand, if the workflow is meant to be use only with public organization members, then it should be able to fetch that list without any credentials using https://api.github.com/orgs/ORG/members endpoint and not require `members.json`.

ian-shafer:
This is a known issue with this workflow that we plan on addressing in the near future. FYI, we would have started with simply accessing the `orgs` REST API, but the token granted to workflows does not allow read access to that part of the REST API.

mattrandallbecker:
I think this is our reasonable first pass as it allows someone to curate the configuration list down to only those that should not require two reviews when making a PR. You can remove any names from it you don't want to be public when creating the configuration.

In the future we'll create a new workflow that only takes a team name as input, and uses [this API](https://docs.github.com/en/rest/teams/members?apiVersion=2022-11-28) to validate the author is within that team. That will obfuscate who is actually in the team, although one could figure it out by seeing which PRs only require one approval. That workflow however will need a PAT token, and so rolling that out will be more complex.

meltsufin:
It's good to know that there are plans to make this easier in the future. For now, I think we'll curate the list manually. Since we have many repositories in our organization, an improvement could be the ability to read the `members.json` from another repository from within the GitHub org.

suztomo:
@ian-shafer  Would you elaborate more about why multi-approvers.yml needs organization membership? The GitHub repository permissions in my team are managed in repository-level by teams, not by organization membership. The people with the "Write" or "Maintain" role carry the risk of leveraging "sock puppet" account.

![Image](https://github.com/user-attachments/assets/61117e95-42ac-432b-98e7-2d3f2a48eeb9)

For example, even though I belong to the googleapis organization, I cannot merge a pull request in https://github.com/googleapis/nodejs-bigtable, which I don't have the roles to merge a pull request by myself.

meltsufin:
We should probably disable non-members of the organization from being added as collaborators to any of the repos.

ian-shafer:
> Would you elaborate more about why multi-approvers.yml needs organization membership?

@suztomo I was probably wrong about that. Based on feedback (including yours) the next step for this workflow is to add support to query team membership at runtime. We could also do this for orgs (as I suggested above), but that does not sound as useful.

[...]

PR #396, opened 2025-02-06
Add support for GitHub teams to multi-approvers

### LINKED ISSUE #364: Multi-approvers workflow to use list of public organization members

ian-shafer:
This change also updates the allowed inputs to the multi-approvers
workflow.

Teams Support
========

Support for GitHub teams requires github-token-minter (Minty) to
generate a token with elevated priviledges to access the team membership
REST APIs.

New Inputs
========

The original org-members-path has been deprecated. It will continue to
be supported in the near future. Users can migrate from org-members-path
by using the new members-url iput.

Three inputs are supported:

1. members-url
2. local-members-path
3. team

See the README for more on each of these inputs.

Fixes #364 and #361


.github/workflows/multi-approvers.yml
```diff
@@ -76,19 +122,55 @@ jobs:
...
+          requested_permissions: |-
+            {
+              "scope": "actions-testing-multi-approvers",
+              "repositories": ["actions-testing"],
+              "permissions": {
+                "members": "read",
+                "pull_requests": "read"
+              }
+            }
+
```
bradegler, line 157: This should be done from the calling workflow. The minty instance referenced from the variables specified here is the `abcxyz` instance and a caller from another org will have no way to change that, rendering this somewhat unusable. 

The current WIF pool is deployed with an assertion that prevents requests from all other orgs which would block your mint token request as well.

Usage of this workflow should look something like:

-> auth
-> minty
-> multi-approvers (passing minty token in)

bradegler (changes requested):

.github/workflows/multi-approvers.yml
ian-shafer: I spent some time digging into this. We're not currently able to pass secrets from one workflow job to another. Some related facts:

The minty token is marked as a secret [here](https://github.com/abcxyz/github-token-minter/blob/58d88c3d9edc3360e927bd23732f4cbfe6eaa6e9/.github/actions/minty/main.js#L45). 

This is actually supported (see [Example: Masking and passing a secret between jobs or workflows](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#example-masking-and-passing-a-secret-between-jobs-or-workflows)), but it's very cumbersome and probably more than we want to require for this. It requires setting up a secrets store like Vault.

We've only used the Minty steps and the generated token within the _same_ job. I was trying to do it across jobs because the multi-approvers workflow is a reusable workflow and therefore it needs to be called in it's own separate job (see [docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_iduses)).

We could look into turning the multi-approvers reusable workflow into an action or multiple actions. I'm not sure if this would work.

Would it be possible to just pass the required values for the two minty actions?

.github/workflows/multi-approvers.js
```diff
@@ -2,49 +2,120 @@ const APPROVED = 'APPROVED';
...
-      const reviewerLogin = r.user.login;
-
-      // Set state if it does not exist.
-      if (!Object.hasOwn(reviewStateByLogin, reviewerLogin)) {
-        reviewStateByLogin[reviewerLogin] = r.state;
-        return;
-      }
+/** Implements approvedCount using the contains method. */
+class AbstractMembers {
+
```
sethvargo, line 7: We don't really write a lot of js so we don't have a linter yet, but there shouldn't be an extra newline here.

.github/workflows/multi-approvers.js
```diff
@@ -2,49 +2,120 @@ const APPROVED = 'APPROVED';
 const COMMENTED = 'COMMENTED';
```
sethvargo, line 2: Is there an option to use typescript here, or does that require some complex compilation step? /cc @verbanicm 

https://nodejs.org/en/learn/typescript/run-natively

.github/workflows/multi-approvers.js
```diff
@@ -2,49 +2,120 @@ const APPROVED = 'APPROVED';
...
+          return;
+        }
+
+        // Do not update approved state for a comment.
+        if (reviewStateByLogin[reviewerLogin] === APPROVED && r.state !== COMMENTED) {
+          reviewStateByLogin[reviewerLogin] = r.state;
+        }
+      })
+
+    return Object.values(reviewStateByLogin).filter((s) => s === APPROVED).length;
```
sethvargo, line 39: This is super weird and would be a lot clearer with a single `for` loop using the quick `continue` pattern.

I also don't really understand why we're putting things into a map, then returning the value? It seems like a lot of unnecessary filtering and iteration.

.github/workflows/multi-approvers.js
```diff
@@ -2,49 +2,120 @@ const APPROVED = 'APPROVED';
...
+    } catch (error) {
+      if (error.status === 404) {
+        // We can get here for two reasons:
+        // 1) The user is not a member
+        // 2) The team does not exist
+        // Either way, it's safe to return false here.
+        return false;
+      }
+      throw error;
+    }
```
sethvargo, line 91: We should really log the error (`this.core.debug`) or something.

.github/workflows/multi-approvers.js
```diff
@@ -2,49 +2,120 @@ const APPROVED = 'APPROVED';
...
-  const members = require(orgMembersPath).reduce((acc, v) => acc.set(v.login, v), new Map());
+async function onPullRequest({orgTeam, membersPath, prNumber, repoName, repoOwner, github, core}) {
+  const members = (function() {
+    if (orgTeam) {
+      if (orgTeam.indexOf('/') <= 0) {
+        throw new Error(`Malformed team [${orgTeam}]. Team must be in the format \${org}/\${team-slug}.`);
+      }
+      const teamOrg = orgTeam.split('/')[0];
+      const teamSlug = orgTeam.split('/')[1];
+      return new TeamMembers(teamOrg, teamSlug, github);
```
sethvargo, line 104: Double split is inefficient. This will also potentially be `undefined` if the `orgTeam` doesn't contain a slash.

.github/workflows/multi-approvers.js
```diff
@@ -2,49 +2,120 @@ const APPROVED = 'APPROVED';
...
+    }
+    if (membersPath) {
+      return new JsonMembers(membersPath);
+    }
+    throw new Error('Neither orgTeam nor membersPath is set.');
```
sethvargo, line 109: What if both are provided? I would expect that to be an error.

.github/workflows/multi-approvers.yml
```diff
@@ -21,7 +21,16 @@ on:
...
+        description: 'DEPRECATED: Use members-url with https://raw.githubusercontent.com/ prepeneded instead.'
+        type: 'string'
+      members-url:
+        description: 'Full URL to JSON members file. URL must be publicly accessible.'
+        type: 'string'
+      local-members-path:
+        description: 'Path to JSON members file. See README for more.'
+        type: 'string'
+      team:
+        description: 'The GitHub org and team slug that should be used for determining external vs. internal users. Format is ${org}/${team-slug}.'
```
sethvargo, line 33: I feel like we really need to simplify these inputs. Why do we need so many?

.github/workflows/multi-approvers.yml
```diff
@@ -40,8 +54,21 @@ jobs:
...
-      - name: 'Download members.json'
-        id: 'download-members-json'
+      - name: 'Check inputs'
+        shell: 'bash'
+        run: |-
+          if [[ -z "${{ inputs.org-members-path }}" &&
+                -z "${{ inputs.members-url }}" &&
+                -z "${{ inputs.local-members-path }}" &&
+                -z "${{ inputs.team }}" ]]; then
+            echo "Invalid inputs. One of org-members-path, members-url, local-members-path, or team must be set." 1>&2
```
sethvargo, line 64: Don't access inputs directly in a script (it's a shell injection vuln). Set inputs as envvars and access the envvars.

.github/workflows/multi-approvers.yml
sethvargo: The caller of the workflow needs to setup/configure Minty and pass in the resulting `GITHUB_TOKEN` to this action.

.github/workflows/multi-approvers.js
verbanicm: This looks like it is experimental and requires a special flag and I dont think we can control this for actions currently. So we would need to do the build process to "compile" into JS from TS.

.github/workflows/multi-approvers.yml
ian-shafer: FYI, Brad and I talked and I'm going to turn the multi-approvers workflow into an action. This will allow the Minty actions and multi-approvers to run in the same job and removes the need to pass a secret across jobs.

README.md
```diff
@@ -205,6 +208,55 @@ gh api \
...
+the workflow.
+
+**members-url**: Full URL to a members JSON file. If this is in a public GitHub
+repository, the URL should be similar to
+https://raw.githubusercontent.com/${owner}/${repo}/${ref}/path/to/members.json.
+
+##### Teams
+
+If you use the `team` input, you will have to setup
+[github-token-minter](https://github.com/abcxyz/github-token-minter)  to provide
```
suztomo, line 234: Am I right about the requirements below?

- The usage of teams here needs github-token-minter.
- github-token-minter needs Terraform (https://github.com/abcxyz/github-token-minter)?

README.md
```diff
@@ -205,6 +208,55 @@ gh api \
...
+team). The github-token-minter scope configuration should be as follows:
+
+```yaml
+scope:
+  your-name: # Set this to whatever you want.
+    rule:
+      if: |-
+        # Check that the request is coming from the expected repository.
+        # The ID should be replaced with yours.
+        assertion.repository_id == '12345678' &&
```
suztomo, line 245: Can we use repository name, such as "google-cloud-java" for https://github.com/googleapis/google-cloud-java? (I've never referenced a GitHub repository with a number. Is there a benefit in this?)

README.md
```diff
@@ -205,6 +208,55 @@ gh api \
...
+concept could also be used for a GitHub team. Of course, the JSON file can be
+generated by any means, as long as it follow the expected format.
+
+##### Inputs
+
+One of `team`, `local-members-path`, or `members-url` must be set. If multiple
+of these inputs are set, only one will be used. Which one will be used is
+undefined, but deterministic.
+
+**team**: This is actually the org and team formateted as ${org}/${team-slug}
```
suztomo, line 221: Can we specify multiple teams? In the repositories we (Cloud SDK) maintain, it's norm to have the service team and the Cloud SDK team maintain a repository together (e.g., [java-storage](https://github.com/googleapis/java-storage))

.github/workflows/multi-approvers.yml
```diff
@@ -21,7 +21,16 @@ on:
...
-        required: true
+        description: 'DEPRECATED: Use members-url with https://raw.githubusercontent.com/ prepeneded instead.'
+        type: 'string'
+      members-url:
+        description: 'Full URL to JSON members file. URL must be publicly accessible.'
+        type: 'string'
+      local-members-path:
+        description: 'Path to JSON members file. See README for more.'
+        type: 'string'
+      team:
```
sakthivelmanii, line 32: Is it possible to configure organization as well? We would like to configure this GitHub action at the organization level. 

@meltsufin @suztomo

README.md
meltsufin: @suztomo We really just need one team of all Googlers who are members of the organization. More granular approvals can be achieved with a CODEOWNERS file where you can specify multiple teams.

README.md
meltsufin: I also see this code though:
```yaml
          # Use the Minty token if it was generated, otherwise use the default.
          github-token: '${{ steps.mint-token.outputs.token || github.token }}'
```

README.md
bradegler: Repository names are mutable **and** subject to name squatting attacks. Using the repository id is the correct way to ensure the request is coming from the repository that it is expected to come from.

README.md
suztomo: Yes, either one organization team, if it exists, suffices the needs.

README.md
suztomo: Great. That's worth mentioning here.

.github/workflows/multi-approvers.js
ian-shafer: Will address this in a different PR where I will use an Action instead of a Workflow.

.github/workflows/multi-approvers.js
ian-shafer: Will address this in a different PR where I will use an Action instead of a Workflow.

.github/workflows/multi-approvers.js
ian-shafer: In the README I said that at least one had to be set and that if multiple are set then an undefined (yet consistent) input would be chosen.

But I agree with you. It'll be less error-prone to only allow one.

I'll address this in a different PR where I will use an Action instead of a Workflow.

.github/workflows/multi-approvers.yml
ian-shafer: I think we need at least two to support GH teams and a members JSON file.

1. GH team descriptor
2. JSON URL

For the JSON URL, it's nice to allow a path to the local repo for manual testing / development / maintenance. This isn't required, but it alleviates the need to find a public place to put a JSON file when testing in a private repo like abcxyz/actions-testsing.

Also note, the JSON members files is nice because it does not require a github-token-minter (Minty) setup.

README.md
ian-shafer: Yes, use of the `team` input will require github-token-minter. You won't need one per repo, but you will need one somewhere. @bradegler can speak more to this.

README.md
ian-shafer: Would it make sense to allow for an `org` input where internal approvers are members of the org instead of a team?

.github/workflows/multi-approvers.yml
ian-shafer: I think I proposed this [above](https://github.com/abcxyz/pkg/pull/396#discussion_r1949464016). I'll chat with my team to see if this is a possibility

ian-shafer:
FYI, I will eventually close this PR in favor of another where we use an Action instead of a Workflow for multi-approvers logic.

README.md
suztomo: Yes, organization membership makes sense to me. It can tell the organization already for the repository URL.

.github/workflows/multi-approvers.yml
sethvargo: I'd really prefer to do that ~now before we have people start depending on the workflow version. I also think, eventually, we'll want to move this into it's own repo like we did with terraform-linter (but that can be a follow-up).

.github/workflows/multi-approvers.js
sethvargo: Please create a follow-up issue.

.github/workflows/multi-approvers.js
sethvargo: Please create a follow-up issue.

.github/workflows/multi-approvers.js
sethvargo: Please create a follow-up issue.

.github/workflows/multi-approvers.js
sethvargo: No, this is a bug that will cause a runtime panic and/or unexpected behavior. It needs to be fixed prior to merging. We do not merge known-bugs into HEAD.

.github/workflows/multi-approvers.js
sethvargo: Please create a follow-up issue. We need exhaustive behavior with unit tests to verify the expected behavior.

.github/workflows/multi-approvers.yml
sethvargo: @sakthivelmanii what is the use case? You cannot assign CODEOWNERS teams outside of the organization, so I don't understand the use case for why you'd want to query teams outside of the org.

.github/workflows/multi-approvers.yml
sethvargo: Why do we need to allow specifying a members JSON at all? That seems useful for local testing, but should never be used in a real scenario. We would never actually want someone manually managing the list of internal members in the organization, since that represents a substantial risk of being out of sync and creating insider risk.

I am **strongly** opposed to complicating a public API surface just for testing. These kinds of things should be unit tests anywhere (where we fake the response from the GitHub API).

.github/workflows/multi-approvers.yml
sethvargo: Please fix before merging, regardless of the future. We cannot have known security exploits sitting in code at HEAD.

sethvargo (changes requested):

.github/workflows/multi-approvers.yml
ian-shafer: I'm sorry I didn't make that clear. This PR will not be merged. I'll close this PR now to make it clear.

RE moving this into its own repo. Why not do that now? I'm not sure how much work it is to setup a new repo, but I'd prefer to do it now to avoid migration costs in the future (all clients will have to migrate to use the new action).

.github/workflows/multi-approvers.yml
ian-shafer: I think @sakthivelmanii is asking for all members of a GH org (/orgs/{org}/members) to be treated as internal reviewers. @sakthivelmanii is this correct?

PR #401, opened 2025-02-19
Add or update ignore files

sethvargo:


.github/actions/create-release/action.yml
```diff
@@ -47,7 +47,7 @@
 #
 #           - uses: 'actions/setup-go@v5'
 #             with:
-#               go-version: '1.21'
+#               go-version:-file 'go.mod'
```
verbanicm, line 50: @sethvargo this looks incorrect

sethvargo: sed miss. Fixed

PR #411, opened 2025-04-01
feat: handle escaped commas in slicevar flags

dcreey:
Allows users to pass `,` values in their string slice items by escaping them.

Example
`-my-slice-var="This is a message\, please consider,Another message"`
Yields
`[]string{"This is a message, please consider", "Another message"}`

### DIFF
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -875,7 +875,7 @@ func (f *FlagSection) StringSliceVar(i *StringSliceVar) {
 		for _, indexPair := range indices {
 			if indexPair[4] != -1 {
 				part := s[lastMatch:indexPair[4]]
-				parsed := strings.TrimSpace(escapeComma(part))
+				parsed := strings.TrimSpace(unescapeCommas(part))
 				if parsed != "" {
 					final = append(final, parsed)
 				}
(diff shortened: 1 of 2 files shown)


verbanicm (approved):

cli/flags.go
```diff
@@ -910,6 +924,10 @@ func (f *FlagSection) StringSliceVar(i *StringSliceVar) {
 	})
 }
 
+func unescapeCommas(v string) string {
+	return strings.ReplaceAll(v, `\,`, ",")
+}
+
```
sethvargo, line 930: I think this is unnecessarily complex over:

```go
parts := strings.Split(s, ",")
for i := len(parts) - 2; i >= 0; i-- {
	if strings.HasSuffix(parts[i], "\\") {
		parts[i] = parts[i][:len(parts[i])-1] + "," + parts[i+1]
		parts = append(parts[:i+1], parts[i+2:]...)
	}
}
```

sethvargo: You could also use `strings.FieldsFunc` or something else to avoid the regex and complex rejoining shown here.

sethvargo (changes requested):

cli/flags.go
dcreey: Updated. Couldn't think of a clean way to use strings.FieldsFunc since it does not execute in guaranteed order.

sethvargo (approved):

PR #413, opened 2025-04-02
cli: permit escaped slashes in values, remove extra iterations

sethvargo:
References https://github.com/abcxyz/pkg/pull/411

### DIFF
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -870,27 +870,23 @@ func (f *FlagSection) StringSliceVar(i *StringSliceVar) {
 		var escaped bool
 
 		for _, r := range s {
-			if r == '\\' {
-				if escaped {
-					b.WriteRune(r)
-					escaped = false
-				} else {
-					escaped = true
-				}
+			if escaped {
+				escaped = false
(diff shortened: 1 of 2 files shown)


cli/flags.go
```diff
@@ -865,14 +865,39 @@ type StringSliceVar struct {
...
+				escaped = false
+				if v := strings.TrimSpace(b.String()); v != "" {
+					parts = append(parts, v)
+				}
+				b.Reset()
+				continue
+			}
+
+			escaped = false
+			b.WriteRune(r)
```
dcreey, line 893: I think this might be a bit cleaner
```suggestion
			if escaped {
				escaped = false
				b.WriteRune(r)
				continue
			}
			switch r {
			case '\\':
				escaped = true
				break
			}
			case ',' {
				if v := strings.TrimSpace(b.String()); v != "" {
					parts = append(parts, v)
				}
				b.Reset()
				break
			default:
				b.WriteRune(r)
				break
			}
```

dcreey (approved):

verbanicm (approved):

PR #415, opened 2025-04-02
feat: allow logging to arbitrary files and stderr

dcreey:
Log messages need to be able to be differentiated from parseable binary output. Optionally allowing users OR cli/binary authors to set a default log target (STDOUT or STDERR) allows for easier downstream parsing.

Use case:
* tagrep: some warning condition is met but does not prevent continuation/printing out tags. The warning message is logged to stdout - which makes the user have to parse logs from the stdout in order to grab the tags.
* guardian: some warning condition is met but does not prevent fetching entrypoints. Now the user has to know how to parse entrypoints from stdout in order to use them.

NOTE: I considered making the default log target stderr - but given that I don't understand all of the usages I'm not sure this makes sense. Instead I will leave it up to each binary author to decide. I will update tagrep to default to stderr log target.

### DIFF
--- a/logging/logger.go
+++ b/logging/logger.go
@@ -117,11 +117,11 @@ func newFromEnv(envPrefix string, getenv func(string) string) *slog.Logger {
 		}
 	}
 
-	logTargetKey := envPrefix + "LOG_TARGET"
-	logTargetValue := strings.TrimSpace(getenv(logTargetKey))
-	target, err := LookupTarget(logTargetValue)
+	targetEnvVarKey := envPrefix + "LOG_TARGET"
+	targetEnvVarValue := strings.TrimSpace(getenv(targetEnvVarKey))
+	target, err := LookupTarget(targetEnvVarValue)
 	if err != nil {
-		panic(fmt.Sprintf("invalid value for %s: %s", logTargetKey, err))
+		panic(fmt.Sprintf("invalid value for %s: %s", targetEnvVarKey, err))
 	}
 
 	return New(target, level, format, debug)
--- a/logging/logger.go
+++ b/logging/logger.go
@@ -117,7 +117,14 @@ func newFromEnv(envPrefix string, getenv func(string) string) *slog.Logger {
 		}
 	}
 
-	return New(os.Stdout, level, format, debug)
+	logTargetKey := envPrefix + "LOG_TARGET"
+	logTargetValue := strings.TrimSpace(getenv(logTargetKey))
+	target, err := LookupTarget(logTargetValue)
+	if err != nil {
+		panic(fmt.Sprintf("invalid value for %s: %s", logTargetKey, err))
+	}
+
+	return New(target, level, format, debug)
 }
 
 // SetLevel adjusts the level on the provided logger. The handler on the given
--- a/logging/logger_test.go
+++ b/logging/logger_test.go
@@ -144,6 +144,12 @@ func TestNewFromEnv(t *testing.T) {
 
 		// target
 		{
+			name: "empty_target",
+			env: map[string]string{
+				"LOG_TARGET": "",
+			},
+		},
+		{
 			name: "custom_target",
 			env: map[string]string{
 				"LOG_TARGET": "STDERR",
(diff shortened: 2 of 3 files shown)


logging/targets.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+	"os"
+	"slices"
+	"strings"
+)
+
+const (
+	targetSTDOUTName = "STDOUT"
```
verbanicm, line 25: nit

```suggestion
targetStdoutName
targetStderrName
```

logging/targets.go
```diff
@@ -0,0 +1,53 @@
...
+	targetSTDERRName = "STDERR"
+)
+
+var targetNames = []string{
+	targetSTDOUTName,
+	targetSTDERRName,
```
verbanicm, line 31: nit

```suggestion
targetStdoutName
targetStderrName
```

dcreey: Done

logging/targets.go
```diff
@@ -0,0 +1,57 @@
...
+
+var targetNames = []string{
+	targetStdoutName,
+	targetStderrName,
+}
+
+// TargetNames returns the list of all log target names.
+func TargetNames() []string {
+	return slices.Clone(targetNames)
+}
```
sethvargo, line 37: I don't think you need this anymore, since you don't enumerate the target options in the `default` case (since default is to fallback to file)

logging/logger.go
```diff
@@ -117,7 +117,14 @@ func newFromEnv(envPrefix string, getenv func(string) string) *slog.Logger {
...
 
-	return New(os.Stdout, level, format, debug)
+	logTargetKey := envPrefix + "LOG_TARGET"
+	logTargetValue := strings.TrimSpace(getenv(logTargetKey))
+	target, err := LookupTarget(logTargetValue)
+	if err != nil {
+		panic(fmt.Sprintf("invalid value for %s: %s", logTargetKey, err))
+	}
+
+	return New(target, level, format, debug)
```
sethvargo, line 127: I did this locally, but there's a pretty bad bug here that's insurmountable: there's no way to hold onto the target file to flush and close it. This isn't required for Stderr/Stdout, but it _is_ required for a real file, or else you'll lose the last few logs written (and you'll leak fds).

sethvargo (changes requested):

logging/logger.go
dcreey: Removed.

logging/logger.go
```diff
@@ -117,7 +117,14 @@ func newFromEnv(envPrefix string, getenv func(string) string) *slog.Logger {
 		}
 	}
 
-	return New(os.Stdout, level, format, debug)
+	logTargetKey := envPrefix + "LOG_TARGET"
+	logTargetValue := strings.TrimSpace(getenv(logTargetKey))
```
sethvargo, line 121: Nit: please use the naming convention above (`EnvVar`)

logging/logger_test.go
```diff
@@ -141,6 +141,21 @@ func TestNewFromEnv(t *testing.T) {
 			},
 			wantPanic: "invalid value for LOG_DEBUG: strconv.ParseBool",
 		},
+
+		// target
```
sethvargo, line 145: Add test for `""` please

logging/targets.go
```diff
@@ -0,0 +1,47 @@
...
+
+// LookupTarget attempts to get the target that corresponds to the given
+// name. If no such target exists, it returns an error. If the empty string
+// is given, it returns the STDOUT target.
+func LookupTarget(name string) (*os.File, error) {
+	switch v := strings.ToUpper(strings.TrimSpace(name)); v {
+	case "":
+		return os.Stdout, nil
+	case targetStdoutName:
+		return os.Stdout, nil
```
sethvargo, line 41:
```suggestion
	case "", targetStdoutName: // "" for backwards-compatibility
		return os.Stdout, nil
```

sethvargo (changes requested):

logging/logger.go
dcreey: Done

logging/logger_test.go
dcreey: Done

PR #416, opened 2025-04-03
feat: global log settings

dcreey:
# Proposal

Add global logging env vars that take precedent over local custom ones.

NOTE: I considered doing the opposite but the problem is that we have established a pattern of setting the custom logging env vars in the CLI/binary - which means that we would never check the global settings.

# Background

As we scale the usage of abc packages for cli and logging we need to provide a consistent and reliable way to set logging configuration across CLIs. This will be useful for googlers wanting to debug a service/binary without having to know the custom prefix utilized for every single service/binary. Imagine an SRE that wants to interact with one of the abc backed services/binaries - they just need a simple way to configure outputs.

### DIFF
--- a/logging/logger.go
+++ b/logging/logger.go
@@ -129,8 +129,12 @@ func newFromEnv(envPrefix string, getenv func(string) string) *slog.Logger {
 		}
 	}
 
-	targetEnvVarKey := envPrefix + "LOG_TARGET"
+	targetEnvVarKey := "LOG_TARGET"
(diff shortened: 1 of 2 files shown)


sethvargo (changes requested):
I'm fine with this change, but I think the logic is inverted. `MYAPP_LOG_LEVEL` should take precedence over `LOG_LEVEL`.

verbanicm (approved):

sethvargo (approved):

PR #418, opened 2025-04-03
logging: simplify multi-lookup

sethvargo:


logging/logger.go
```diff
@@ -82,62 +82,54 @@ func New(w io.Writer, level slog.Level, format Format, debug bool) *slog.Logger
...
+// from the environment. It sources the following environment variables, first
+// checking any with the prefix, then falling back to the global unprefixed
+// value:
 //
 //   - LOG_LEVEL: string representation of the log level. It panics if no such log level exists.
 //   - LOG_FORMAT: format in which to output logs (e.g. json, text). It panics if no such format exists.
 //   - LOG_DEBUG: enable the most detailed debug logging. It panics iff the given value is not a valid boolean.
 func NewFromEnv(envPrefix string) *slog.Logger {
-	return newFromEnv(envPrefix, os.Getenv)
+	return newFromEnv(envPrefix, func(ss ...string) (string, string) {
```
dcreey, line 93: NIT: I feel like this function def should not be inlined. There is no clear reason and its a bit harder to read.

dcreey (approved):

dcreey (approved):

verbanicm (approved):

PR #419, opened 2025-04-03
chore: create lint actions to replace logic from lint workflows

dcreey:
### What

New lint actions with the inputs and steps copied from the existing lint workflows.

### Why

1. GH Enterprise Server does not support referencing reusable workflows from github.com (even with github connect) - so if we want to leverage the same shared code it must be a shared action.
3. The strategy for [auto linting](https://docs.google.com/document/d/1nJ4wKycxev8d3-FaMwiP4a3YMMVmseofJIPwbwElIaA/edit?resourcekey=0-OYMYvuqjCVlcEbRpvq4O7w&tab=t.0#heading=h.gqscjct0tgq1) on all repositories will require using reusable actions for each linter instead of reusable workflows. This is because we don't know which linters to run without detecting which files are present in a repo - then we utilize github actions matrix syntax to dynamically run each linter.

### Follow Ups

1. Update the reusable workflows for linting to use these new actions
2. Create the new general purpose lint.yml workflow that may replace the reusable workflows.

### DIFF
--- a/.github/actions/lint-go/action.yml
+++ b/.github/actions/lint-go/action.yml
@@ -43,16 +43,45 @@ runs:
       with:
         go-version: '${{ inputs.go_version }}'
         go-version-file: '${{ inputs.go_version_file }}'
+        cache: false
 
-    - name: 'Check modules'
-      shell: 'bash'
-      working-directory: '${{ inputs.directory }}'
+    - name: 'Lint (download default configuration)'
+      id: 'load-default-config'
+      if: |-
+        ${{ hashFiles('.golangci.yml', '.golangci.yaml') == '' }}
       run: |-
-        for d in $(find . -name go.mod); do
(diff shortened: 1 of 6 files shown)


.github/actions/lint-go/action.yml
verbanicm, line 1: I think this file should be lint-go-modules and vice versa

verbanicm (approved):
See comment re: naming

.github/actions/lint-go/action.yml
dcreey: Good catch

dcreey:
> Is there a way we can reduce the duplication? We have mostly the same thing in another file, and keeping them in sync will be difficult.

I'm going to call these actions from the workflows in order to dedupe - I just can't make the change in the same PR. Its a follow up task

PR #422, opened 2025-04-04
chore: update workflows to use new actions

dcreey:


.github/workflows/go-lint.yml
```diff
@@ -54,24 +54,14 @@ jobs:
       - name: 'Checkout'
         uses: 'actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683' # ratchet:actions/checkout@v4
 
-      - name: 'Setup Go'
-        uses: 'actions/setup-go@f111f3307d8850f501ac008e886eec1fd1932a34' # ratchet:actions/setup-go@v5
+      - name: 'Lint Go Modules'
+        uses: 'abcxyz/pkg/.github/actions/lint-go-modules@main' # ratchet:exclude
```
sethvargo, line 58: Should we use the relative paths (`../actions/lint-go-modules`)? The current approach requires committing changes to main before these actions pick them up.

dcreey: I don't think relative paths don't work for when its called from another repository.

dcreey: One workaround is that we create local lint-test workflows that use relative paths. It creates some duplication tho in terms of checks.

dcreey: Sorry, accidentally closed this PR - moving to https://github.com/abcxyz/pkg/pull/424

PR #424, opened 2025-04-04
chore: update workflows to use new actions

dcreey:


.github/workflows/go-lint.yml
```diff
@@ -54,24 +54,14 @@ jobs:
       - name: 'Checkout'
         uses: 'actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683' # ratchet:actions/checkout@v4
 
-      - name: 'Setup Go'
-        uses: 'actions/setup-go@f111f3307d8850f501ac008e886eec1fd1932a34' # ratchet:actions/setup-go@v5
+      - name: 'Lint Go Modules'
+        uses: 'abcxyz/pkg/.github/actions/lint-go-modules@main' # ratchet:exclude
```
sethvargo, line 58: From the other thread:

CleanShot 2025-04-04 at 14 09 27@2x

I'm really worried that we're now "one commit" removed from all CI in `pkg`, which feels bad. It means we could break a workflow, but not find out about it until the next CI run post-merge to `main`.

I don't follow your point about relative paths and reusability. With the new architecture, people should be using the stuff in `actions/` anyway, right?

sethvargo (changes requested):

.github/workflows/go-lint.yml
dcreey: Its my understanding that reusable workflows stored in a repo that has not been cloned (in the context of running in another repo's workflow) do not have access to resource via relative path in the reusable workflow's repository. This is because it does not clone the workflow's context - it just takes the content of the workflow and executes it in the other repo's context.

sethvargo: Can we test that? I think there's real downsides to having us one-commit removed from changes.

dcreey: Yes, lets test it. The workaround is to have a workflow that references all of the local relative paths just for this repo. That way we test the current commit too.

dcreey: Tested - it fails. https://github.com/abcxyz/actions-testing/pull/90

.github/workflows/ci.yml
```diff
@@ -27,19 +27,49 @@ concurrency:
 
 jobs:
   go_lint:
-    uses: './.github/workflows/go-lint.yml'
+    runs-on: 'ubuntu-latest'
+    steps:
+      - name: 'Checkout'
+        uses: 'actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683' # ratchet:actions/checkout@v4
```
sethvargo, line 33: Do you need to specify which repo to checkout?

sethvargo, line 33: We may want to do a sparse checkout as well as set a different path to avoid conflicts with a normal checkout of the same name.

https://github.com/actions/checkout?tab=readme-ov-file#fetch-only-the-root-files-and-github-and-src-folder

sethvargo (approved):

.github/workflows/ci.yml
dcreey: Nope. When this is run via a reusable workflow it will clone the calling repo's content.

dcreey: I don't think we need to worry about that. This workflow will not be used as a reusable workflow (since it has the relative paths). So this is the only clone that will happen. The workflow that will get called from other workflows will look pretty different. I'm going to create that in a follow up

PR #425, opened 2025-04-07
chore: use tagrep in want-lgtm=all

dcreey:
Updates the want_lgtm=all check to use tagrep. This allows us to easily support running this check in the merge queue in addition to the pull_request context.

WANT_LGTM=all


verbanicm (approved):

dcreey (commented):
Adding a review to test the pull_request_review check

PR #426, opened 2025-04-07
feat: linter for all repos

dcreey:
Proof of concept linter workflow that we can require on all repositories. 

* Initially I thought we could just use the `on::paths` attribute to trigger jobs. But this is not sufficient as this will create the condition that only files matching these paths that have _changed_ in the PR trigger these workflows. We want to trigger these workflows if the repo has any file matching these paths. Otherwise we cannot make the lint jobs required status checks since they will not reliably run on every PR.
* The init workflow checks for files have been tested against a shallow checkout of the chromium repo. Its performant (~1 sec to scan)
* We have a duplicate lint-pkg.yml workflow that only runs in this context. It is generated via a template and if any changes are made to the lint.yml we fail unless you regenerate. This is tested: https://github.com/abcxyz/pkg/actions/runs/14365896307/job/40278605726?pr=426
* TODO: Support reading workflow lint configuration from a file in the root of the repo.
* TODO: We will need to update the required checks again.
* TODO: Create a new .golangci.yml without specific import ordering

WANT_LGTM=all


.github/workflows/lint.yml
```diff
@@ -0,0 +1,283 @@
...
+      ##############
+      # The version of Go to install and use.
+      GO_LINT_GO_VERSION: |-
+        ${{ vars.GO_LINT_GO_VERSION }}
+      # Path to the go.mod file to extract a version.
+      GO_LINT_GO_VERSION_FILE: |-
+        ${{ vars.GO_LINT_GO_VERSION_FILE || 'go.mod' }}
+      # The URL to a golangci file. This is only used if no file is found in the local directory.
+      GO_LINT_GOLANGCI_URL: |-
+        ${{ vars.GO_LINT_GOLANGCI_URL || 'https://raw.githubusercontent.com/abcxyz/pkg/main/.golangci.yml' }}
```
bradegler, line 92: Maybe we should have a more generic `.golangci.yml` that is the default that doesn't force ordering of our packages? That one specifically orders `github.com/abcxyz` packages. It doesn't hurt anything, but might feel weird to someone external that uses our packages.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,283 @@
...
+      matrix:
+        lint-target: '${{ fromJSON(needs.init.outputs.lint-targets) }}'
+    # TODO: Consider parsing from config file in the repo.
+    env:
+      ##############
+      ## GOLANG ####
+      ##############
+      # The version of Go to install and use.
+      GO_LINT_GO_VERSION: |-
+        ${{ vars.GO_LINT_GO_VERSION }}
```
bradegler, line 86: Is the thought process that these variables will be set at the org or repo level?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,283 @@
...
+          go_version: '${{ inputs.GO_LINT_GO_VERSION }}'
+          go_version_file: '${{ inputs.GO_LINT_GO_VERSION_FILE }}'
+          golangci_url: '${{ inputs.GO_LINT_GOLANGCI_URL }}'
+          directory: '${{ inputs.GO_LINT_DIRECTORY }}'
+          golangci_lint_version: '${{ inputs.GO_LINT_GOLANGCI_LINT_VERSION }}'
+
+      - name: '[GOLANG] Lint Go abcxyz/pkg'
+        if: |
+          matrix.lint-target == 'golang' && github.repository == 'abcxyz/pkg'
+        uses: './.github/actions/lint-go'
```
bradegler, line 159: Hmm. I kind of feel like we should always lint against @main and not against the current version in the PR. That would simplify this to not require the special case for `abcxyz/pkg`.

What is the benefit that you see from running the version in the PR? My assumption is that it allows you to test linter changes, but is there anything else?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,283 @@
...
+        uses: 'actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332' # ratchet:actions/checkout@v4
+        with:
+          fetch-depth: 1 # shallow clone
+
+      - name: '[GOLANG] Lint Go'
+        if: |
+          matrix.lint-target == 'golang' && github.repository != 'abcxyz/pkg'
+        uses: 'abcxyz/pkg/.github/actions/lint-go@main'
+        with:
+          go_version: '${{ inputs.GO_LINT_GO_VERSION }}'
```
bradegler, line 150: Should these be `inputs`? Shouldn't they be `env`?

.github/workflows/lint.yml
dcreey: I think the goal is that the user can set their own configuration - I was thinking that we could read a config file from the repo level but also default to vars set at the org or repo level.

.github/workflows/lint.yml
dcreey: Just testing linter changes so that we are not n-1 away on every PR. I put some more info in the PR description

.github/workflows/lint.yml
dcreey: env. Nice catch

.github/workflows/lint.yml
dcreey: I'll do this in a follow up

bradegler (approved):

.github/workflows/lint.yml
dcreey: Updated.

verbanicm (approved):

PR #432, opened 2025-04-11
feat: load lint config from file

dcreey:


dcreey:
> Why are we introducing a config file?

The goal is to allow repositories to override some configuration in an easy manner. Right now they can do it with vars but that is not transparent as only maintainers and above can see those and they also are not code reviewed.

For example: the repo should be able to specify which directory to target for each linter, certain directories to ignore, what versions of golang, java, terraform, typescript, etc. I also think it makes sense to allow repos to easily turn on and off the default linters.

sethvargo:
I'm really worried we're going down the path of super-linter, which was explicitly a non-goal.

PR #437, opened 2025-04-16
chore: deprecate reusable actions

dcreey:
Deprecates all actions except for scancode which needs to be updated in actions repo


.github/actions/create-release/action.yml
```diff
@@ -90,72 +90,14 @@ outputs:
...
-          core.setOutput('version', version)
-
-          // Set the tag (which has the leading "v") prefix.
-          const tag = `v${version}`;
-          core.info(`👾 Computed tag as: "${tag}"`)
-          core.setOutput('tag', tag)
+    - name: 'Emit deprecation warning'
+      shell: 'bash'
+      run: |-
+        echo "::warning::This action has moved to its own repo and will be removed in a future release. Please upgrade to https://github.com/abcxyz/actions... instead."
```
sethvargo, line 96: I realize this is toilsome, but can we go the extra mile and provide the exact name for each of these?

dcreey: Done

sethvargo (approved):

PR #443, opened 2025-04-17
feat: add a reusable workflow to check if PR has description

sailorlqh:
This PR add a new workflow that can be triggered by `workflow_call`. It will fail if the description is empty.


sethvargo:
@dcreey can you sync with @sailorlqh on doing this in the main linter instead

sethvargo (changes requested):
Can you chat with @dcreey about putting this in abcxyz/actions as a default linter instead?

sailorlqh:
Closing this PR and move discussion with @dcreey offline.

Issue #323, opened 2024-04-09
Add exec wrapper library


drevell:
### TL;DR

Both [guardian](https://github.com/abcxyz/guardian/blob/b01fa398bb7ca21a4c0402ee1be902af9a7d5a45/pkg/child/child.go#L56) and [abc](https://github.com/abcxyz/abc/blob/49babab96513de693386e38b4ca1ff6a335aeeb1/templates/common/run/run.go#L62) have helper libraries that wrap the os/exec library. In https://github.com/abcxyz/abc/pull/531#discussion_r1556014743 we agreed that these probably should be factored out into a shared common library in pkg.

PR #445, opened 2025-04-21
[WIP -- looking for feedback] Unified Exec Wrapper

pdewilde:
This is my first crack at #323:
---
Both [guardian](https://github.com/abcxyz/guardian/blob/b01fa398bb7ca21a4c0402ee1be902af9a7d5a45/pkg/child/child.go#L56) and [abc](https://github.com/abcxyz/abc/blob/49babab96513de693386e38b4ca1ff6a335aeeb1/templates/common/run/run.go#L62) have helper libraries that wrap the os/exec library. In https://github.com/abcxyz/abc/pull/531#discussion_r1556014743 we agreed that these probably should be factored out into a shared common library in pkg.
---

I have tried to take the features of both and make something combined, though I fear maybe this is getting too messy.

Changes:

- Removed the `Many()` that the abc library had. It feels like it won't be often used and logic may better live in the client, as they may not want the `Simple()` semantics.
- Removed the ProcessAttributes and cancel changes from Guardian. These should be added back in, but I didn't feel Guardian implementation was great and didn't want to overcomplicate this PR. (Guardian only special-cases Linux, but all unix deployment paths should support signaling).

Concerns:
- This does seem to be getting a bit unwieldy
- Does this actually add enough value to stdlib to warrant existing?


run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/abcxyz/pkg/logging" // Import the specific logging package
+)
+
+// DefaultRunTimeout is how long commands wait if the context doesn't have a deadline.
+const DefaultRunTimeout = time.Minute
```
sailorlqh, line 36: RunTimeout and wait delay have different units, do you think it might be more user friendly to change this to `60 * time.Second`?

pdewilde: The are both time.Duration which under the hood is in nanoseconds.

They are actually just
```
const (
	Nanosecond  [Duration](https://pkg.go.dev/time#Duration) = 1
	Microsecond          = 1000 * [Nanosecond](https://pkg.go.dev/time#Nanosecond)
	Millisecond          = 1000 * [Microsecond](https://pkg.go.dev/time#Microsecond)
	Second               = 1000 * [Millisecond](https://pkg.go.dev/time#Millisecond)
	Minute               = 60 * [Second](https://pkg.go.dev/time#Second)
	Hour                 = 60 * [Minute](https://pkg.go.dev/time#Minute)
)
```

IDK if 60 * time.Second is easier to grok than time.Mintue

sethvargo: They are all int64 under the hood so it doesn't matter. I do wonder why we're setting such a long default? It seems like most commands should finish in like < 5s, or manually have no timeout. We also want the ability to disable the timeout entirely, since we rely on the subcommand to handle the timeouts.

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+	logger.DebugContext(ctx, "set wait delay", "delay", cmd.WaitDelay)
+
+	logger.DebugContext(ctx, "starting command",
+		"args", args,
+		"options", compiledOpts,
+	)
+	err := cmd.Run() // This blocks until the command exits or context is cancelled
+
+	// Exit code -1 if ProcessState is nil (e.g., start failed)
+	exitCode = -1
```
sethvargo, line 146: I don't think shells allow exit codes to be negative.

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	return exitCode, err
+}
+
+// Option implements the functional options pattern for [Run].
+// It holds all configurable settings for executing a command.
+type Option struct {
+	// Basic execution control
+	allowNonZeroExit bool
```
sethvargo, line 202: We should be more generic with "allowed exit codes"

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+	err := cmd.Run() // This blocks until the command exits or context is cancelled
+
+	// Exit code -1 if ProcessState is nil (e.g., start failed)
+	exitCode = -1
+	if cmd.ProcessState != nil {
+		exitCode = cmd.ProcessState.ExitCode()
+	}
+
+	if err != nil {
+		var exitErr *exec.ExitError
```
sethvargo, line 152: It's possible for the `ExitError` to include an exit status:

```go
if exiterr, ok := err.(*exec.ExitError); ok {
	if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
		code = status.ExitStatus()
	}
}
```

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+			err = nil
+		} else if isExitError {
+			stdoutContent := "[stdout not captured]"
+			if stdoutBuf != nil {
+				stdoutContent = stdoutBuf.String()
+			}
+			stderrContent := "[stderr not captured]"
+			if stderrBuf != nil {
+				stderrContent = stderrBuf.String()
+			}
```
sethvargo, line 166: If these are empty, we should leave them as empty

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+// Option implements the functional options pattern for [Run].
+// It holds all configurable settings for executing a command.
+type Option struct {
+	// Basic execution control
+	allowNonZeroExit bool
+	cwd              string
+	stdin            io.Reader
+	stdout           io.Writer
+	stderr           io.Writer
+	waitDelay        *time.Duration
```
sethvargo, line 207: This shouldn't be a pointer ([it's an int64](https://pkg.go.dev/time#Duration))

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+// standard input.
+func WithStdin(stdin io.Reader) *Option {
+	return &Option{stdin: stdin}
+}
+
+// WithStdinStr is a convenience option that uses the given string as the
+// command's standard input.
+func WithStdinStr(stdin string) *Option {
+	return WithStdin(bytes.NewBufferString(stdin))
+}
```
sethvargo, line 230: `strings.NewReader("my string")` is really succinct. I don't think we should expose this surface.

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+// TODO: handle process group and cancel semantics.
+// WithProcessGroup returns an option that attempts to run the command in a new
+// process group (primarily effective on Unix-like systems).
+//func WithProcessGroup(set bool) *Option {
+//	return &Option{setProcessGroup: set}
+//}
+
+// The last option specified for most fields takes precedence.
+// additionalEnv is appended across options.
+func compileOpts(opts []*Option) *Option {
```
sethvargo, line 279: I think we want the "struct pattern" instead of the "options pattern" for this. It will make documentation and possible arguments more intuitive. We are already passing around an options structure to keep track of everything anyway.

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+			finalEnv = append(finalEnv, v)
+		}
+	}
+
+	// Add overrides at the end, after any filtering. os/exec has "last wins"
+	// semantics, so appending is an overriding action.
+	finalEnv = append(finalEnv, overrideEnv...)
+
+	return finalEnv
+}
```
sethvargo, line 349: I'd rather expose this functionality publicly and simplify the interface to only accept the final `[]string` environ.

run/exec.go
pdewilde: Hmm, that seems fine, but runs to the question: what if you want to allow all exit codes? Do we need a separate flag? Do we accept a magic value that means all? Do we accept a slice of strings so we can have some wildcard?

run/exec.go
pdewilde: Yes, but sometimes there isn't actually an exit code (executable not found, or interrupted before could exit)

os uses this same convention, defaulting to -1
https://pkg.go.dev/os#ProcessState.ExitCode

run/exec.go
pdewilde: https://github.com/abcxyz/abc/blob/49babab96513de693386e38b4ca1ff6a335aeeb1/templates/common/run/run.go#L30 

I kept the default from `abc`. I assume the thinking there was it likely would be calling a long-running process (compilation or something similar).

Maybe for a general library that should be shorter though

run/exec.go
sethvargo: I typed a reply to this, but idk where it went... I think we can accept a slice of allowed exit codes. If the slice is empty, we expect 0. Otherwise we only accept the codes in the list.

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+
+	cmd.Dir = compiledOpts.cwd
+	cmd.Stdin = compiledOpts.stdin
+	// Handle stdout/stderr, capturing output if writers are bytes.Buffer for error reporting.
+	var stdoutBuf, stderrBuf *bytes.Buffer
+	if compiledOpts.stdout != nil {
+		cmd.Stdout = compiledOpts.stdout
+		if bb, ok := compiledOpts.stdout.(*bytes.Buffer); ok {
+			stdoutBuf = bb
+		}
```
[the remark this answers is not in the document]
pdewilde: Yeah, that is weird.

I was using gemini to combine some of the logic, cleaned up most of the cruft but seems this crept through. I think I see what it was going at.

Original code:

```go
err = fmt.Errorf(`exec of %v failed: error was "%w", context error was "%w"\nstdout: %s\nstderr: %s`, args, err, ctx.Err(), cmd.Stdout, cmd.Stderr)
```

Since bytes.Buffer is a stringer, this meant that in the case of bytes.Buffer the stderr and stdout would be printed in the error message. Simple used this and documented the behavior in the godoc. My guess is gemini wanted to be helpful and default to [stderr/out not captured]  in the other case, rather than giving something like a FD. I would personally prefer to default to the stringer as before, so i'll revert that.

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+	useCustomEnv := len(compiledOpts.allowedEnvKeys) > 0 || len(compiledOpts.deniedEnvKeys) > 0 || len(compiledOpts.additionalEnv) > 0
+	if useCustomEnv {
+		currentEnv := os.Environ()
+		logger.DebugContext(ctx, "calculating custom environment",
+			"inherited_count", len(currentEnv),
+			"allowed_patterns", compiledOpts.allowedEnvKeys,
+			"denied_patterns", compiledOpts.deniedEnvKeys,
+			"additional_vars_count", len(compiledOpts.additionalEnv))
+		env := environ(currentEnv, compiledOpts.allowedEnvKeys, compiledOpts.deniedEnvKeys, compiledOpts.additionalEnv)
+		logger.DebugContext(ctx, "computed environment", "env", env)
```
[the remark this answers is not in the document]
pdewilde: Removed the env logic as suggested, so no longer applicable

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+	if compiledOpts.stderr != nil {
+		cmd.Stderr = compiledOpts.stderr
+		if bb, ok := compiledOpts.stderr.(*bytes.Buffer); ok {
+			stderrBuf = bb
+		}
+	} else {
+		cmd.Stderr = os.Stderr
+	}
+
+	useCustomEnv := len(compiledOpts.allowedEnvKeys) > 0 || len(compiledOpts.deniedEnvKeys) > 0 || len(compiledOpts.additionalEnv) > 0
```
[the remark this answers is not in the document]
pdewilde: Good point, changed

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+	} else {
+		cmd.WaitDelay = DefaultWaitDelay
+	}
+	logger.DebugContext(ctx, "set wait delay", "delay", cmd.WaitDelay)
+
+	logger.DebugContext(ctx, "starting command",
+		"args", args,
+		"options", compiledOpts,
+	)
+	err := cmd.Run() // This blocks until the command exits or context is cancelled
```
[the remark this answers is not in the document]
pdewilde: Because its using a commandcontext, canceling the context will send a stop signal.

As far as blocking goes, I think that should be fine, if someone really cares they can always wrap in a goroutine.

I could see an argument if you feel like there will likely need to be other signals sent to the command, in which case it would be nice to get the PID back somehow.

run/exec.go
pdewilde: Hmm, what about int[]{-1} representing "accept any exit code"

run/exec.go
pdewilde: This has all been removed.

run/exec.go
pdewilde: Its a pointer as its an optional value where the default value has semantic meaning (don't wait at all) different from the default behaivor we want (set to DefaultWaitDelay const)

run/exec.go
pdewilde: Yeah, removed

run/exec.go
pdewilde: Done

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+// Simple is a wrapper around [Run] that captures stdout and stderr as strings.
+// This is intended to be used for commands that run non-interactively then
+// exit.
+//
+// If the command exits with a nonzero status code, an *exec.ExitError will be
+// returned.
+//
+// If the command fails, the error message will include the contents of stdout
+// and stderr. This saves boilerplate in the caller.
+func Simple(ctx context.Context, args ...string) (stdout, stderr string, _ error) {
```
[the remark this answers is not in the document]
pdewilde: Yeah, not sure if there is a better way to do this

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+
+	if err != nil {
+		var exitErr *exec.ExitError
+		isExitError := errors.As(err, &exitErr)
+
+		if isExitError && compiledOpts.allowNonZeroExit {
+			logger.DebugContext(ctx, "command exited non-zero, but allowed by option", "exit_code", exitCode)
+			err = nil
+		} else if isExitError {
+			stdoutContent := "[stdout not captured]"
```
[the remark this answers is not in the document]
pdewilde: Cleaned up a bit

run/exec.go
sethvargo: The default value for an int (and time.Duration) is 0, so it's functionally equivalent? Or are you trying to differentiate between "0" and "unset"?

run/exec.go
sethvargo: I don't think there's any use case for "allow any exit code"?

run/exec.go
```diff
@@ -0,0 +1,349 @@
...
+// DefaultWaitDelay is how long after context cancellation processes are actually killed
+// if another value isn't set. See exec.Cmd.WaitDelay for more information.
+const DefaultWaitDelay = time.Second
+
+// Simple is a wrapper around [Run] that captures stdout and stderr as strings.
+// This is intended to be used for commands that run non-interactively then
+// exit.
+//
+// If the command exits with a nonzero status code, an *exec.ExitError will be
+// returned.
```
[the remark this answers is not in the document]
pdewilde: Couldn't you just cast the string back to a []byte if that was an issue?

My understanding is go strings are just []byte under the hood, and are not required to hold valid utf-8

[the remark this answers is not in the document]
pdewilde: https://go.dev/play/p/9nCBm5wgPaq

run/exec.go
pdewilde: https://github.com/abcxyz/abc/blob/main/templates/common/run/diff.go#L85

Its being used by abc currently to simplify error handling (you don't need to cast the error to check if its exit code vs some other type)

run/exec.go
pdewilde: Done

run/exec.go
pdewilde: Yes. Default value for int* is nil, which will cause `DefaultWaitDelay` to be used.

0 is already well defiend by cmd.Run:
```
	// If WaitDelay is zero (the default), I/O pipes will be read until EOF,
	// which might not occur until orphaned subprocesses of the command have
	// also closed their descriptors for the pipes.
```

which isn't the default behavior in guardian's library, assumed that behavior was wanted: https://github.com/abcxyz/guardian/blob/main/pkg/child/child.go#L93

run/exec.go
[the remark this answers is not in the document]
sethvargo: Casting copies the data...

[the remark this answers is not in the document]
pdewilde: How about this:

```
// Simple is a wrapper around [Run] that captures stdout and stderr as strings.
// This is intended to be used for commands that run non-interactively then
// exit. For large amounts of output, it's recommended to call Run with a
// streaming writer to avoid having all output in RAM.
```

I feel like in either case (binary or string output) if its large enough to care about copies, probably good idea to stream output anyways.

